PR #15, opened 2023-04-24
renegade_contracts: nullifier_set: Implement `NullifierSet` contract in Cairo 1.0

akirillo:
Implementation of our `NullifierSet` contract in Cairo 1.0, at feature and test parity with our Cairo 0 implementation.

This avoids the need for any scripting / "integration" tests because we can test the logic directly within Cairo 1.0!

Note: the nullifiers are stored as `u256` instead of `felt252` because they'll be Dalek scalar field elements.


Cargo.lock
```diff
@@ -327,9 +327,9 @@ dependencies = [
 
 [[package]]
```
joeykraut: do we need to commit the lock file? These things tend to bloat repos

src/nullifier_set/tests.cairo
```diff
@@ -0,0 +1,58 @@
...
+
+fn is_nullifier_used_helper(nullifier: u256) -> bool {
+    let mut nullifier_used = NullifierSet::__external::is_nullifier_used(
+        test_utils::serialized_element(nullifier)
+    );
+    test_utils::single_deserialize(ref nullifier_used)
+}
+
+// Achieves the same thing as the Cairo 0 `test_valid_nullifiers` test,
+// minus the bit of fuzzing done there (selecting 20 random nullifiers).
```
joeykraut, line 18: nit: Let's not reference code that's going to be deleted down the line, would prefer the comment just describe what we're testing

joeykraut (approved):
lgtm, nice work, super clean interfaces!

joeykraut:
> lgtm, nice work, super clean interfaces!

Lastly, can we clean up the commit history via rebase? I'd prefer to have a clean history that describes an actionable step taken at each commit. I.e. no "fmt" commits, and add scoping. This may seem small now, but it will greatly help us when we need to retro bugs

Cargo.lock
akirillo: oh ack my b

src/nullifier_set/tests.cairo
akirillo: ack, though coming back to this comment I can actually implement this now, will do so

PR #16, opened 2023-04-25
renegade_contracts: merkle: Implement `Merkle` contract in Cairo 1.0

akirillo:
This PR implements our `Merkle` contract in Cairo 1.0, mirroring our Cairo 0 implementation as closely as possible.

A few places where it diverts, though:
- Using `u128` for the indices, and `u8` for the height instead of `felt252`, mostly for readability
- Using Quaireaux's `fast_power` [implementation](https://github.com/keep-starknet-strange/quaireaux/blob/4c93251c89409313edcee8f9dce64416cb38d9c9/quaireaux/math/src/fast_power.cairo#L15) 
- Caching the tree capacity in contract storage so it is not recalculated on each update of the `next_index`

Note: The intention is to have the Merkle nodes be `u256` instead of `felt252`, but for now I've left them this way to make testing against an arkworks implementation simpler. The change will have to happen once I implement Poseidon hashing over the Dalek scalar field in Cairo 1.0.

This is also almost at test parity with the Cairo0 contracts, namely we test against an `arkworks` impl of a Merkle tree (using Starknet's Pedersen hash) to assert:
- Correct root, history on initialization
- Correct root, history after single insert
- Correct root, history after multiple inserts
- Fails when trying to insert to an already full tree

We're still missing tests that events are emitted properly. While this is possible, it's quite janky (have to issue RPC requests against the local devnet node), and I held off on this (and general code organization / over-abstraction of the test suite) because I'm not yet committed to our testing toolstack.

At the moment, we use [`nile-rs`](https://github.com/OpenZeppelin/nile-rs) to run (Rust) scripts which invoke the `nile-rs` CLI to declare/deploy/invoke contracts against a [`starknet-devnet`](https://github.com/0xSpaceShard/starknet-devnet) node running locally. However, [`starknet_in_rust`](https://github.com/lambdaclass/starknet_in_rust) is almost done testing & documenting their Cairo 1 support, and I want to see how that goes before committing too hard either way. `starknet_in_rust` could be really nice because we could interact with Starknet state w/o running a devnet node (and issuing RPC calls under the hood), and could interact with it directly in Rust. If this works well I'd also gladly rip `nile-rs` out of our stack for now, it's a pretty poor dev experience writing within their scripting framework (they do a custom build script that inlines your script to a `main.rs`, this mucks up `rust-analyzer` which makes me sad)


akirillo:
Ah, our testing CI fails because `cairo-test` tries to compile the project from scratch, and doesn't pull in the dependencies (unlike compiling with `scarb build`).

It seems implementing a Cairo test runner is pretty low priority for the Scarb team right now: https://github.com/software-mansion/scarb/issues/222

I've asked for workarounds in the Scarb channel on the Starknet discord, but if there are none, our options are:
1. Vendor the Quaireaux functionality we use (`fast_power` and `check_gas`) into the repo
2. Roll our own naive implementations of said functionality

akirillo:
The Scarb folks got back to me - for now, we can use [scarb-eject](https://github.com/software-mansion-labs/scarb-eject) to vendor in the deps. I've made the appropriate CI changes in the Cairo1.0 main branch and rebased this onto it

Cargo.lock
```diff
@@ -108,21 +108,76 @@ version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
```
joeykraut: Same as in the other PR, don't think we need to commit the lock file

scripts/.module/src/merkle.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    type LeafHash = FeltCRH;
+    type TwoToOneHash = FeltTwoToOneCRH;
+    type LeafInnerDigestConverter = IdentityDigestConverter<[u8; 32]>;
+}
+
+pub type FeltMerkleTree = MerkleTree<MerkleConfig>;
+
+pub const EMPTY_LEAF_VAL: &str =
+    "306932273398430716639340090025251549301604242969558673011416862133942957551";
```
joeykraut, line 81: What is this constant and how did we get to it? If it's the same one as in the existing implementation can we also borrow the comment explaining what this is?

scripts/test_merkle.rs
```diff
@@ -0,0 +1,283 @@
+use eyre::{eyre, Result};
+use starknet_crypto::FieldElement;
+use num_bigint::{BigUint, RandBigInt};
+
+// Requires a devnet node running
+async fn run(_nre: NileRuntimeEnvironment) -> Result<()> {
+
+    let contract_name = "Merkle";
+    let merkle_test_builder = MerkleTestBuilder::default().with_contract_name(contract_name.to_string()).with_height(5);
```
joeykraut: nit: Did we run `rustfmt` on all the `.rs` files? I'd suggest making this part of CI too, this line looks long for example, but maybe this is just how `rustfmt` likes it?

scripts/test_merkle.rs
```diff
@@ -0,0 +1,283 @@
...
+// - The contract state is uninitialized
+
+#[derive(Default)]
+struct MerkleTestBuilder {
+    contract_name: Option<String>,
+    contract_address: Option<String>,
+    height: Option<usize>,
+}
+
+impl MerkleTestBuilder {
```
joeykraut: Do we need the builder pattern here? Seems like a constructor on `MerkleTest` would do fine? I'm not very opinionated on this so feel free to ignore, but it seems like we just error anyways if these fields aren't set.

scripts/test_merkle.rs
```diff
@@ -0,0 +1,283 @@
...
+        Ok(())
+    }
+
+    fn get_roots(&self) -> Result<(FieldElement, FieldElement)> {
+        debug!("Getting root from arkworks Merkle tree...");
+        let arkworks_root = FieldElement::from_bytes_be(&self.merkle_tree.root())?;
+        debug!("Got root: {arkworks_root:#?}");
+
+        debug!("Getting root from Merkle contract...");
+        let contract_root = utils::call(&self.contract_address, "get_root", vec![])?[0];
```
joeykraut: nit: I'd make `"get_root"` et al into constants defined at the module level

scripts/test_merkle.rs
```diff
@@ -0,0 +1,283 @@
...
+struct MerkleTest {
+    contract_name: String,
+    contract_address: String,
+    height: usize,
+    merkle_tree: merkle::FeltMerkleTree,
+    next_index: usize,
+}
+
+impl MerkleTest {
+
```
joeykraut, line 112: Use `#![allow(non_snake_case)]` to disable this lint at the block level, or alternatively do
```
#[allow(non_snake_case)]
impl MerkleTest {
    ...
}
```

scripts/test_merkle.rs
```diff
@@ -0,0 +1,283 @@
...
+        debug!("Inserting {leaf_val} into Merkle contract...");
+        utils::send(&self.contract_address, "insert", vec![&leaf_val.to_string()])
+    }
+
+    fn insert_val_to_arkworks(&mut self, leaf_val: BigUint) -> Result<()> {
+        debug!("Inserting {leaf_val} into arkworks Merkle tree...");
+        let mut leaf_val_bytes: [u8; 32] = [0; 32];
+        let leaf_val_bytes_vec = leaf_val.to_bytes_be();
+        // Unset bits pushed to beginning of array b/c big-endian
+        leaf_val_bytes[32 - leaf_val_bytes_vec.len()..].copy_from_slice(&leaf_val_bytes_vec);
```
joeykraut: Bit unclear what's happening here from the current implementation. It seems like what we're trying to do is pad the `BigUint::to_bytes_be` up to size 32. In this case I think a more clear and idiomatic (albeit verbose) way to do this is with an iterator:
```rust
let mut bytes_be = val_bigint
        .to_bytes_le()
        .into_iter()
        .chain(iter::repeat(0u8))
        .take(32)
        .collect_vec()
    bytes_be.reverse();
```
Using the [itertools](https://docs.rs/itertools/latest/itertools/) crate.

We don't have to go this route, but if you choose to leave as-is, please update the comment to clarify a bit

src/merkle.cairo
```diff
@@ -0,0 +1,32 @@
...
+    // | INTERFACE |
+    // -------------
+
+    #[external]
+    fn initializer(height: u8) {
+        MerkleLib::initializer(height);
+    }
+
+    #[view]
+    fn get_root() -> felt252 {
```
joeykraut, line 18: For transparency & to confirm: the plan is to change this return type to some value that supports the Dalek field size, correct?

src/merkle/library.cairo
```diff
@@ -0,0 +1,263 @@
...
+    /// The index represents the index into the sibling array; i,e, the list formed by
+    /// reading the nodes at the given height, left to right.
+    #[event]
+    fn Merkle_internal_node_changed(height: u8, index: u128, new_value: felt252) {}
+
+    // -----------
+    // | LIBRARY |
+    // -----------
+
+    const MAX_CAPACITY: u128 = 340282366920938463463374607431768211455_u128;
```
joeykraut, line 68: nit: add a comment on this magic number

src/merkle/library.cairo
```diff
@@ -0,0 +1,263 @@
...
+        let height = height::read();
+        let next_index = next_index::read();
+        let tree_capacity = capacity::read();
+
+        assert(next_index < tree_capacity, 'merkle tree full');
+
+        // Increment the next index if the tree has room
+        next_index::write(next_index + 1_u128);
+        return next_index;
+    }
```
joeykraut: nit: In terms of function ordering, might read nicer if we grouped all the helpers, or at least put them below the first function that calls them

src/merkle/library.cairo
```diff
@@ -0,0 +1,263 @@
...
+    /// Insert a value into the Merkle tree
+    /// Parameters:
+    /// - `value`: The value to insert into the tree
+    /// Returns:
+    /// - The root computed by hashing the value into the tree
+    fn insert(value: felt252) -> felt252 {
+        let height = height::read();
+
+        // Increment the index of the next empty leaf
+        let next_index = _increment_next_index();
```
joeykraut: nit: rename to `curr_index`

src/merkle/library.cairo
```diff
@@ -0,0 +1,263 @@
...
+        let (next_index, is_right) = u128_safe_divmod(insert_index, u128_as_non_zero(2_u128));
+        let is_left = (is_right == 0_u128);
+
+        // If the subtree rooted at the current node is filled, update the sibling value
+        // for the next insertion. There are two cases here:
+        //      1. The current insertion index is a left child; in this case the updated
+        //         sibling value is the newly computed node value.
+        //      2. The current insertion index is a right child; in this case, the subtree
+        //         of the parent is filled as well, meaning we should set the updated sibling
+        //         to the zero value at this height; representing the parent's right child
```
joeykraut: I know I wrote this comment haha, but let's change it to something like "representing the parent's next empty sibling"

joeykraut (approved):
Overall lgtm, just a few small thing, but nothing blocking.

Can we also clean up the commit history to be a bit more precise and actionable?

scripts/test_merkle.rs
akirillo: yeahhhh fair enough

scripts/.module/src/merkle.rs
akirillo: ack good call

scripts/test_merkle.rs
akirillo: ah, yeah i'll add to CI, but this file not being part of the package as specified by `Cargo.toml` makes it not get picked up in the formatting 🙃
I'm going to hold off on setting this up to work until I can assess `nile-rs`+`starknet-devnet` vs `starknet_in_rust`, if we go with the latter (which I suspect we will) then we'll have a normal project structure without having to do whatever whacky fix that also points `rustfmt` at this

scripts/test_merkle.rs
akirillo: this is hot

src/merkle.cairo
akirillo: yes, absolutely - holding off on this until after I've implemented poseidon hashing in cairo, as it was much simpler to test the merkle impl over felts

src/merkle/library.cairo
akirillo: ack i like this

PR #17, opened 2023-06-01
renegade_contracts: merkle, nullifier_set, testing: darkpool updates

akirillo:
This PR makes the necessary changes to the merkle & nullifier set contracts, as well as to the Cairo 1-native testing structure, for full integration of these contracts into the final darkpool contract.

To be explicit: The complete Cairo1 rewrite is on the Darkpool-Cairo1.0 branch, this branch is one of many in which I try to split up those changes to make reviews more piecemeal


scripts/.module/build.rs
```diff
@@ -19,9 +19,10 @@ fn main() {
         with_disclosure
             + &r#"
 pub mod utils;
-pub mod merkle;
 pub mod nullifier_set;
-use tracing::log::{debug, error};
+pub mod merkle;
+// pub mod darkpool;
```
joeykraut, line 24: Dead code?

scripts/.module/src/main.rs
```diff
@@ -4,20 +4,38 @@ use eyre::Result;
...
+    //     erc20_contract_address: String::from("0x0274408801d9030031611a934f58c281e01d0fd99435535d955404d583e5438c"),
+    //     darkpool_class_hash: String::from("0x0517ed8ea5ad37f835323070b675f6f6110f82780b3e5201fcc2ad1bdc21f643"),
+    //     darkpool_contract_address: String::from("0x0216aac50b0dc047133c0d5d5f01f38b72dee6112e5b95365591779fe1941c78"),
+    //     merkle_class_hash: String::from("0x031b1da91bf3c92387adf9fd1ce106464b6c26990155b7c36e36d32237ff06e2"),
+    //     merkle_contract_address: String::from("0x0116fefb6fb8dcbf29955037927503eaa0cc6d7fa6f8b996dd3fe52c240d3493"),
+    //     nullifier_set_class_hash: String::from("0x02d3b2088f1b805566b10e52eb1153c2c4e76d2371af0c187d0721fc0c08cdce"),
+    //     nullifier_set_contract_address: String::from("0x052a68140ea05601c3d6314621b9b21ae02ac8cdb2d4dafc2a615a146c6c867a"),
+    //     upgrade_target_class_hash: String::from("0x01a48b6acbe06c23ec6cf2355f0fd91103510113cbd575b7753209e0818fdc96"),
+    //     account_address: String::from("0x07e00d496e324876bbc8531f2d9a82bf154d1a04a50218ee74cdd372f75a551a"),
+    // });
```
joeykraut: Dead code?

scripts/.module/src/main.rs
```diff
@@ -31,6 +49,8 @@ fn main() {
 
         let mut devnet = utils::devnet_utils::spawn_devnet().await;
         let res = run().await;
+        // If we panic on an assertion in a test, we never get here...
+        // Maybe just don't use asserts?
         debug!("Killing devnet...");
```
joeykraut, line 45: If you can afford to not use futures or ensure that your futures are `Send` and `Sync`; then you can use `catch_unwind` ([docs](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html)) here to isolate and handle panics in tests. Just a thought, no change explicitly requested :)

scripts/.module/src/merkle/tests.rs
```diff
@@ -201,68 +208,46 @@ impl MerkleTest {
...
-        debug!("Got root: {arkworks_root:#?}");
-
-        debug!("Getting root from {} contract...", &self.contract_name);
-        let contract_root = devnet_utils::call(
-            self.contract_address.clone(),
-            GET_ROOT_FN_NAME.to_string(),
-            vec![],
-        )?[0];
-        debug!("Got root: {contract_root:#?}");
+        let arkworks_root = get_ark_root(&self.merkle_tree)?;
```
joeykraut, line 211: It probably makes sense to have `get_ark_root` as a method of `MerkleTest`. Seems like we wouldn't really need an arkworks implementation outside of the Merkle tests

scripts/.module/src/utils/common_utils.rs
```diff
@@ -26,3 +164,235 @@ pub fn gen_random_felt(bitsize: u64) -> Result<FieldElement> {
...
+            darkpool_class_hash,
+            darkpool_contract_address,
+            merkle_class_hash,
+            merkle_contract_address,
+            nullifier_set_class_hash,
+            nullifier_set_contract_address,
+            upgrade_target_class_hash,
+            account_address,
+        } = dump;
+    } else {
```
joeykraut: This business of declaring variables then assigning them via a destructuring is not very idiomatic or readable. I'd suggest either
```
let state = if let Some(some_dump) = dump { some_dump } else { ... };
```
And then access the variables as `state.account_address`, etc.

Or you can do
```
let state = dump.unwrap_or_else(|| {
    <state fetch logic>
});
```

scripts/.module/src/utils/common_utils.rs
```diff
@@ -26,3 +164,235 @@ pub fn gen_random_felt(bitsize: u64) -> Result<FieldElement> {
...
+        .unwrap();
+    NULLIFIER_SET_CLASS_HASH
+        .set(nullifier_set_class_hash)
+        .unwrap();
+    NULLIFIER_SET_CONTRACT_ADDRESS
+        .set(nullifier_set_contract_address)
+        .unwrap();
+    UPGRADE_TARGET_CLASS_HASH
+        .set(upgrade_target_class_hash)
+        .unwrap();
```
joeykraut: This could also all be a `lazy_static` right? Since it's a `OnceCell` anyways; seems like access could be restructured to avoid this `Option` unwrapping logic.

src/lib.cairo
```diff
@@ -1,4 +1,8 @@
-mod test_utils;
+mod testing;
 mod utils;
-mod merkle;
 mod nullifier_set;
+mod merkle;
+// mod darkpool;
+// mod oz;
```
joeykraut, line 5: I'm getting the sense this is commented out because the next PR follows up with these modules? If this is the case ignore my previous comments about dead code

src/nullifier_set/library.cairo
```diff
@@ -29,19 +30,21 @@ mod NullifierSetLib {
...
-    fn mark_nullifier_used(nullifier: u256) {
+    fn mark_nullifier_used(nullifier: felt252) {
         // Assert that the nullifier hasn't already been used
         assert(!nullifier_spent_set::read(nullifier), 'nullifier already spent');
 
         // Assert nullifier is in Dalek scalar field
-        assert(nullifier < dalek_order(), 'nullifier not in field');
+        // For now, this is a trivial check, as nullifiers are felt252s
+        // and necessarily fit into the Dalek scalar field order
+        assert(nullifier.into() < dalek_order(), 'nullifier not in field');
```
joeykraut: do we need this if it's always going to be true?

src/testing/tests/nullifier_set_tests.cairo
```diff
@@ -0,0 +1,54 @@
...
+}
+// #[test]
+// #[available_gas(300000)]
+// #[should_panic]
+// fn test_nullifier_not_in_field() {
+//     let nullifier = dalek_order() + 1.into();
+
+//     // Mark nullifier as used
+//     NullifierSet::__external::mark_nullifier_used(test_utils::serialized_element(nullifier));
+// }
```
joeykraut: We can just remove this, I don't see us doing non-native

joeykraut (approved):
lgtm; some smaller suggestions, nothing blocking

scripts/.module/build.rs
akirillo: coming soon ™️ in #18

scripts/.module/src/main.rs
akirillo: ugh... this is part of something I jerry-rigged together to speed up testing, declaring/deploying the contracts takes a really long time (like 10-15min) so I dump/load state, and when I do so I just cache the class hashes / contract addresses I need in this struct. for cleanliness I can just remove this from the repo and only use it for local testing

src/nullifier_set/library.cairo
akirillo: ah yup good call, this was back when we were expecting to pivot to `u256` for the nullifiers, but it seems like we're pretty committed to using the STARK curve, so will rip this stuff out

scripts/.module/src/main.rs
akirillo: did a bit of poking about and it looks like `UnwindSafe` can't be inferred for the futures I await throughout the testing code, since I'm not committed to this testing structure anyway I don't think it's worth grinding through at the moment

scripts/.module/src/merkle/tests.rs
akirillo: ah, we end up needing this functionality in the darkpool tests too, which is why I have it in the `common_utils` file

scripts/.module/src/main.rs
joeykraut: all good, it's fine to leave in at your judgement

scripts/.module/src/utils/common_utils.rs
akirillo: hmm, I'm not sure - if we use a `lazy_static` block in this function, the scope of the static variables is limited and they can't be imported. if we try to declare the variables in a `lazy_static` block at the top, then we need to wrap them all in mutexes so they can be set in this function, which is equally unwieldy IMO. It feels like a big lift to have all the function logic in a `lazy_static` at the top level since I set the variables w/ the results of async calls...

PR #18, opened 2023-06-02
renegade_contracts: darkpool: initial darkpool implementation

akirillo:
This PR introduces the full first version of the darkpool contract, with testing, access controls, and upgradeability.

A note on the test scripts: they are in quite a sloppy state at the moment. This stems from a few optimistic initial decisions on what tool stack to use, which caused me to entrench myself in the `nile-rs` scripting patterns while learning more and more how far from ideal they were.

The end state is that I am invoking `nile-rs` CLI from Rust against a local devnet node and doing a lot of `stdout` parsing, because their Rust scripting runtime is far from feature-complete. We decided it was not worth the time to pivot to another tooling stack (I now err on the side of Protostar + `starknet_in_rust`), and as such my testing utilities got increasingly hacky... the core logic of the tests is sound, but the code is not pretty. This is certainly a refactor I look forward to handling down the line.


scripts/.module/src/main.rs
```diff
@@ -4,35 +4,53 @@ use eyre::Result;
...
+            "0x052a68140ea05601c3d6314621b9b21ae02ac8cdb2d4dafc2a615a146c6c867a",
+        ),
+        upgrade_target_class_hash: String::from(
+            "0x01a48b6acbe06c23ec6cf2355f0fd91103510113cbd575b7753209e0818fdc96",
+        ),
+        account_address: String::from(
+            "0x07e00d496e324876bbc8531f2d9a82bf154d1a04a50218ee74cdd372f75a551a",
+        ),
+    });
+    // let dump = None;
```
joeykraut: dead code?

src/oz/ownable/library.cairo
```diff
@@ -0,0 +1,48 @@
...
+
+    fn initializer(owner: ContractAddress) {
+        _transfer_ownership(owner);
+    }
+
+    fn assert_only_owner() {
+        let owner: ContractAddress = _owner::read();
+        let caller: ContractAddress = get_caller_address();
+        assert(!caller.is_zero(), 'Caller is the zero address');
+        assert(caller == owner, 'Caller is not the owner');
```
joeykraut, line 25: Out of curiosity is there a situation in which the first assert would fail where the second does not? I.e. what does the first assertion add? I like that it's here for better error handling, but curious if there's any other reason

src/testing/test_contracts/dummy_erc20.cairo
```diff
@@ -0,0 +1,128 @@
...
+    fn decrease_allowance(spender: ContractAddress, subtracted_value: u256) {
+        let caller = get_caller_address();
+        approve_helper(caller, spender, allowances::read((caller, spender)) - subtracted_value);
+    }
+
+    fn transfer_helper(sender: ContractAddress, recipient: ContractAddress, amount: u256) {
+        assert(!sender.is_zero(), 'ERC20: transfer from 0');
+        assert(!recipient.is_zero(), 'ERC20: transfer to 0');
+        balances::write(sender, balances::read(sender) - amount);
+        balances::write(recipient, balances::read(recipient) + amount);
```
joeykraut, line 110: nit: This is for testing so it doesn't matter that much but can we add a comment indicating that this contract is not safe for deployment due to lack of range checks?

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+        /// in which it was changed
+        wallet_last_modified: LegacyMap<felt252, felt252>
+    }
+
+    // ----------
+    // | EVENTS |
+    // ----------
+
+    /// Emitted when there's an update to the encrypted wallet identified by `public_blinder_share`
+    // TODO: Do we want this at all? Why are we doxxing the wallet that was updated?
```
joeykraut: Yes, this is useful not just for consumers but for searching through our state transition history down the line!

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+    // ---------
+
+    /// Initializes the contract state
+    /// Parameters:
+    /// - `merkle_class`: The declared contract class of the Merkle tree implementation
+    /// - `nullifier_class`: The declared contract class of the nullifier set implementation
+    fn initializer(
+        _merkle_class_hash: ClassHash, _nullifier_set_class_hash: ClassHash, _height: u8
+    ) {
+        // TODO: Add `proxy_admin` param & Proxy init logic
```
joeykraut: Is this TODO still relevant? Seems like the caller handles this. I think we should structure all authorization to happen in the caller and these methods assume correct auth.

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+        private_share_commitment: felt252,
+        public_wallet_shares: Array::<felt252>,
+        proof_blob: Array::<felt252>,
+    ) -> felt252 {
+        // TODO: Add verification of `VALID_WALLET_CREATE`
+
+        // Insert the new wallet's commitment into the Merkle tree
+        let merkle_tree = _get_merkle_tree();
+        merkle_tree.insert(public_share_commitment);
+        let new_root = merkle_tree.insert(private_share_commitment);
```
joeykraut: These two are committed to together; note the changes after the [encryption redesign](https://github.com/renegade-fi/renegade-contracts/blob/joey/encryption-interface-update/contracts/darkpool/library.cairo#L229). I think those changes are at a maturity to which it makes sense to copy them here.

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+        public_wallet_shares: Array::<felt252>,
+        mut external_transfers: Array::<ExternalTransfer>,
+        proof_blob: Array::<felt252>,
+    ) -> felt252 {
+        // TODO: Add verification of `VALID_WALLET_UPDATE`
+
+        // Insert the updated wallet's commitment into the Merkle tree
+        let merkle_tree = _get_merkle_tree();
+        merkle_tree.insert(public_share_commitment);
+        let new_root = merkle_tree.insert(private_share_commitment);
```
joeykraut: Same as above and everywhere else in the file, I'd just copy the changes over from that branch

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+    ) -> felt252 {
+        let nullifier_set = _get_nullifier_set();
+        nullifier_set.mark_nullifier_used(party_0_payload.old_shares_nullifier);
+        nullifier_set.mark_nullifier_used(party_1_payload.old_shares_nullifier);
+
+        let merkle_tree = _get_merkle_tree();
+        merkle_tree.insert(party_0_payload.public_share_commitment);
+        merkle_tree.insert(party_0_payload.private_share_commitment);
+        merkle_tree.insert(party_1_payload.public_share_commitment);
+        let new_root = merkle_tree.insert(party_1_payload.private_share_commitment);
```
joeykraut: same as above

src/darkpool/library.cairo
```diff
@@ -0,0 +1,416 @@
...
+        new_root
+    }
+
+    // -----------
+    // | HELPERS |
+    // -----------
+
+    // What am I doing... can't I just call functions on the libraries directly instead of doing library calls?
+    // Ah - this is in case we want to upgrade the implementations. We don't have to redeploy the Darkpool,
+    // just re-declare the libraries.
```
joeykraut: nit: remove this

scripts/.module/src/darkpool/tests.rs
```diff
@@ -0,0 +1,930 @@
...
+        old_shares_nullifier
+    )?);
+
+    devnet_utils::load_devnet_state().await
+}
+
+async fn test_update_wallet__deposit() -> Result<()> {
+    // Account 0 is blessed w/ funds from test init
+    // Deposit into darkpool and check both balances
+    init_darkpool(0, false)?;
```
joeykraut: nit: for magic constants like this add an inline comment describing what they are, i.e.:
```
init_darkpool(0 /* arg_name */, false /* arg_name */);
```

scripts/.module/src/darkpool/tests.rs
```diff
@@ -0,0 +1,930 @@
...
+        mint: FieldElement::from_hex_be(get_once_cell_string(&ERC20_CONTRACT_ADDRESS)?)?,
+        amount: [FieldElement::from(100u8), FieldElement::ZERO],
+        is_deposit: FieldElement::ONE,
+    }];
+
+    update_wallet(
+        gen_random_felt(MAX_FELT_BIT_SIZE)?,
+        gen_random_felt(MAX_FELT_BIT_SIZE)?,
+        gen_random_felt(MAX_FELT_BIT_SIZE)?,
+        gen_random_felt(MAX_FELT_BIT_SIZE)?,
```
joeykraut: nit: add inlines here too, helps with readability which in turn gives auditability

scripts/.module/src/darkpool/tests.rs
joeykraut: and elsewhere where this is done

scripts/.module/src/darkpool/tests.rs
```diff
@@ -0,0 +1,930 @@
...
+        public_share_commitment,
+        private_share_commitment,
+        Vec::new(),
+        Vec::new(),
+    )?;
+
+    let mut ark_merkle_tree = init_arkworks_merkle_tree(MERKLE_HEIGHT);
+    insert_val_to_arkworks(&mut ark_merkle_tree, 0, public_share_commitment)?;
+    insert_val_to_arkworks(&mut ark_merkle_tree, 1, private_share_commitment)?;
```
joeykraut: Do we need to create a new wallet first? I know we will eventually, but if we can avoid mocking data we should to improve readability

joeykraut (commented):
Overall looks good, when we port over the changes that merge the public and private commitments, we should be good to go!

src/oz/ownable/library.cairo
akirillo: to be honest chief idk, i ripped this directly from OpenZeppelin

they have plans to make themselves available on Scarb (the Cairo package manager) once they stabilize a release, at which point I'll remove this forked code from the repo

scripts/.module/src/main.rs
akirillo: uhhh just bad code tbh
i'll remove this, it's my hacky way of caching declared/deployed contract artifacts

src/darkpool/library.cairo
akirillo: ack, will remove

src/testing/test_contracts/dummy_erc20.cairo
akirillo: (fwiw though, the cairo integer types have implicit range checks, i.e. the `+` / `-` etc. operators are overloaded to be checked arithmetic operations)

src/darkpool/library.cairo
akirillo: LOL

PR #19, opened 2023-06-17
renegade_contracts: utils: add generic utils needed in verification

akirillo:
This PR introduces some of the utilities needed in queueing a verification job.

I decided to put these in a top-level `utils` module since they are generic/agnostic of any verifier types, and I will later refactor the existing contracts to use them where necessary.

Expect CI to be red as it has not been upgraded to target `v2.0.0-rc2` of the Cairo compiler.

Although it's not in this PR for brevity, these utils have seen basic testing in the `initialize` and `queue_verification_job` tests for the verifier (coming soon to a PR near you).

### DIFF
--- a/src/utils/collections.cairo
+++ b/src/utils/collections.cairo
@@ -1,0 +1,65 @@
+use option::OptionTrait;
+use array::ArrayTrait;
+use array::SpanTrait;
+
+
+trait DeepSpan<T, TSpan> {
+    fn deep_span(self: @Array<T>) -> Span<TSpan>;
+}
+
+impl ArrayDeepSpanImpl<T> of DeepSpan<T, T> {
+    fn deep_span(self: @Array<T>) -> Span<T> {
+        self.span()
+    }
+}
+
+impl NestedArrayDeepSpanImpl<T> of DeepSpan<Array<T>, Span<T>> {
+    fn deep_span(self: @Array<Array<T>>) -> Span<Span<T>> {
+        let mut spans = ArrayTrait::new();
+        let mut top_span = self.span();
+        loop {
+            match top_span.pop_front() {
+                Option::Some(span) => spans.append(span.deep_span()),
+                Option::None(()) => {
+                    break;
+                },
+            };
+        };
+        spans.span()
+    }
+}
+
+fn fill_felt_arr(ref arr: Array::<felt252>, val: felt252, mut len: usize) {
+    loop {
+        if len == 0 {
+            break;
+        } else {
+            arr.append(val);
+            len -= 1;
+        };
+    }
+}
+
+fn extend<T, impl TDrop: Drop<T>>(ref arr1: Array::<T>, mut arr2: Array::<T>) {
+    match arr2.pop_front() {
+        Option::Some(v) => {
+            arr1.append(v);
+            extend(ref arr1, arr2);
+        },
+        Option::None(()) => (),
+    }
+}
+
+// FORKED FROM ALEXANDRIA, MADE COMPATIBLE W/ V2.0.0-RC2:
+#[generate_trait]
+impl ArrayImpl<T, impl TDrop: Drop<T>> of ArrayTraitExt<T> {
+    fn append_all(ref self: Array<T>, ref arr: Array<T>) {
+        match arr.pop_front() {
+            Option::Some(v) => {
+                self.append(v);
+                self.append_all(ref arr);
+            },
+            Option::None(()) => (),
+        }
+    }
+}
--- a/src/utils/collections.cairo
+++ b/src/utils/collections.cairo
@@ -29,7 +29,7 @@ impl NestedArrayDeepSpanImpl<T> of DeepSpan<Array<T>, Span<T>> {
     }
 }
 
-fn fill_felt_arr(ref arr: Array::<felt252>, val: felt252, mut len: usize) {
+fn tile_felt_arr(ref arr: Array::<felt252>, val: felt252, mut len: usize) {
     loop {
         if len == 0 {
             break;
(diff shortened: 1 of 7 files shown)


src/utils/collections.cairo
```diff
@@ -0,0 +1,65 @@
...
+                Option::None(()) => {
+                    break;
+                },
+            };
+        };
+        spans.span()
+    }
+}
+
+fn fill_felt_arr(ref arr: Array::<felt252>, val: felt252, mut len: usize) {
```
joeykraut: nit: maybe `tile_felt_arr`?

src/utils/eq.cairo
```diff
@@ -0,0 +1,97 @@
...
+impl EcPointPartialEq of PartialEq<EcPoint> {
+    fn eq(lhs: @EcPoint, rhs: @EcPoint) -> bool {
+        // Serializing EcPoint => obtaining x, y coords (felt252s)
+        // Comparing equality of felt coords is an appropriate notion of equality for EcPoints
+        let mut lhs_ser = ArrayTrait::new();
+        lhs.serialize(ref lhs_ser);
+        let mut rhs_ser = ArrayTrait::new();
+        rhs.serialize(ref rhs_ser);
+
+        lhs_ser == rhs_ser
```
joeykraut: Why not compare the EC points directly? This feels a bit indirect and i don't love the idea of trying a eq operation to an underlying serialization operation.

src/utils/math.cairo
```diff
@@ -0,0 +1,91 @@
+use option::OptionTrait;
```
joeykraut, line 1: Can we add some in-cairo tests for these utils? My impression of our discussion was that we can still do Cairo unit tests given the testing stack issues

src/utils/math.cairo
```diff
@@ -0,0 +1,91 @@
...
+        };
+    };
+    result
+}
+
+// FORKED FROM ALEXANDRIA, MADE COMPATIBLE W/ V2.0.0-RC2:
+fn fast_power(mut base: u128, mut power: u128, modulus: u128) -> u128 {
+    // Return invalid input error
+    if base == 0 {
+        panic_with_felt252('II')
```
joeykraut, line 60: what's the `"II"` here? can we have a more descriptive error message if that's what this represents

src/utils/storage.cairo
```diff
@@ -0,0 +1,146 @@
...
+    storage_access::{
+        StorageBaseAddress, storage_address_from_base, storage_address_from_base_and_offset,
+    },
+    syscalls::{storage_read_syscall, storage_write_syscall}
+};
+
+use super::{collections::{ArrayTraitExt}, serde::EcPointSerde};
+
+
+impl TStorageAccess<
```
joeykraut: nit: Add a doc comment for why this impl exists, it's less clear than the others

src/utils/storage.cairo
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// We use this wrapper trait around Serde (which prepends a length to the serialized data)
+// so that we can do a blanket implementation of StorageAccess for types that impl StorageSerde.
+// If we were to do a blanket implementation for types that impl Serde, we'd have conflicting
+// StorageAccess implementations for some types.
+trait StorageSerde<T> {
+    fn serialize_storage(self: @T, ref output: Array<felt252>);
+    fn deserialize_storage(ref serialized: Span<felt252>) -> Option<T>;
+}
```
joeykraut: This seems to really just be a wrapper to connect serde and storage access traits. I don't think a trait wrapper is the right call here because it adds an odd layer of indirection via a blanket implementation.

Can we instead use a wrapper struct and define a direct implementation of `StorageAccess` on it? I.e.
```
struct StorageAccessSerdeWrapper<T> {
    inner: T
}

impl SerdeImpl<T, impl TSerde: Serde<T>> of StorageAccess<T> {
    ...
}
```
This seems a more direct way of providing blanket functionality

joeykraut (approved):
Some comments, but I'll approve optimistically

src/utils/eq.cairo
akirillo: The `EcPoint` type doesn't have `PartialEq` implemented on it, so they can't be compared directly. Comparing their serialization does feel weird ig, but really what it's doing is comparing their x,y coords, which feels pretty direct

src/utils/math.cairo
akirillo: I'm happy to add unit tests for these, but maybe in a future PR...
Right now I want to get some sort of a complete verifier done to make sure the design holds water.
I have basic confidence in the impl of these functions because they're tested indirectly in `initialize` and `queue_verification_job` tests I introduce in #21 & #22

src/utils/math.cairo
akirillo: I copied this code, alongside the `ArrayTraitExt` trait, from https://github.com/keep-starknet-strange/alexandria - the reason being that I can't use them as a Scarb dependency yet given they haven't updated the code to target the v2.0.0-rc2 compiler, since it's not yet stable.

As such I want to minimize the diff on this code, knowing with high confidence that it will soon leave our codebase in favor of their updated dependencies

src/utils/storage.cairo
akirillo: i think i see where you're coming from, lemme take a stab at it

src/utils/eq.cairo
joeykraut: It's an insidious coupling of two library functions, it may make sense now, but this sort of things is a footgun as the codebase grows. Why not just compare `x` and `y` directly using an `ec_point_unwrap` as you do in the serde impl

akirillo: mm, point taken - will do so

PR #20, opened 2023-06-17
renegade_contracts: verifier: add verifier types & utils

akirillo:
This PR introduces the core types associated with the verifier, as well as some verifier-specific utilities.

As before, expect CI to be red, and know that these types/utils are consumed in `initialize` / `queue_verification_job` tests.

EDIT:
After having started on the implementation of step verification, some of these types / utils have changed. These changes will be reflected in a future PR, but I'm mentioning it now to indicate the review on this PR can be slightly more cursory.

### DIFF
--- a/src/verifier/types.cairo
+++ b/src/verifier/types.cairo
@@ -1,0 +1,202 @@
+use option::{OptionTrait, OptionSerde};
+use array::ArrayTrait;
+use array::SpanTrait;
+use ec::EcPoint;
+
+use renegade_contracts::utils::serde::{EcPointSerde};
+use renegade_contracts::utils::eq::{
+    EcPointPartialEq, ArrayTPartialEq, OptionTPartialEq, TupleSize2PartialEq
+};
+
+
+/// Tracks the verification of a single proof
+#[derive(Drop, Serde, PartialEq)]
+struct VerificationJob {
+    /// The challenge scalars remaining to be used for verification
+    // TODO: This may just have to be a pointer (StorageBaseAddress) to the array
+    scalars_rem: Array<VecPoly3>,
+    /// Tracks the powers of challenge scalar y left to compute
+    y_powers_rem: RemainingScalarPowers,
+    /// Tracks the powers of challenge scalar z left to compute
+    z_powers_rem: RemainingScalarPowers,
+    /// Tracks the G generators left to sample from the hash chain
+    G_rem: RemainingGenerators,
+    /// Tracks the H generators left to sample from the hash chain
+    H_rem: RemainingGenerators,
+    /// The proof-specific commitments remaining to be used for verification
+    commitments_rem: Array<EcPoint>,
+    /// The accumulated result of the verification MSM
+    msm_result: Option<EcPoint>,
+    // The final verdict of the verification
+    verified: Option<bool>,
+}
+
+/// Represents a polynomial (sum of terms) whose evaluation is a scalar used
+/// for a single point in the verification MSM
+type VecPoly3 = Array<VecPoly3Term>;
+
+/// Represents a single term of the scalar polynomial above.
+///
+/// The particular structure of a term is derived from what is actually
+/// required for the verification MSM. The highest-degree terms in the MSM
+/// are of degree 3, and consistently have the form:
+///
+/// scalar * some power of y * some element of a flattened weight, s, or s inv vector
+///
+/// A polynomial containing these terms is evaluated over increasing powers of y & indices
+/// of the vector, and is multiplied across a slice of generators (i.e., it is a sub-MSM)
+///
+/// There are other degree-3 terms in the MSM that instead have the form:
+///
+/// scalar * scalar * scalar
+///
+/// However, these terms are only evaluated once & multiplied by a single generator,
+/// so we just evaluate them up-front and store the result in the `scalar` field here.
+#[derive(Drop, Serde, PartialEq)]
+struct VecPoly3Term {
+    /// The scalar multiple
+    scalar: Option<felt252>,
+    /// Whether or not this term uses a power of y. If so, then the power is calculated
+    /// & tracked in the `y_powers_rem` field of the `VerificationJob` struct.
+    /// When the `num_exponentiations` field of the `y_powers_rem` struct is 0, then
+    /// the polynomial containing this term is exhaused and can be dropped, moving on to the next sub-MSM
+    uses_y_power: bool,
+    /// The vector element to multiply by. When the next index is equal to the max index,
+    /// the polynomial containing this term is exhausted and can be dropped, moving on to the next sub-MSM
+    vec_elem: Option<VecElem>,
+}
+
+#[generate_trait]
+impl VecPoly3Impl of VecPoly3Trait {
+    fn new() -> VecPoly3 {
+        ArrayTrait::new()
+    }
+
+    fn add_term(
+        self: VecPoly3, scalar: Option<felt252>, uses_y_power: bool, vec_elem: Option<VecElem>
+    ) -> VecPoly3 {
+        let mut terms = self;
+        terms.append(VecPoly3Term { scalar, uses_y_power, vec_elem,  });
+        terms
+    }
+
+    fn single_scalar_poly(scalar: felt252) -> VecPoly3 {
+        VecPoly3Trait::new()
+            .add_term(scalar: Option::Some(scalar), uses_y_power: false, vec_elem: Option::None(()))
+    }
+
+    fn map_to_single_scalar_polys(ref scalars: Span<felt252>) -> Array<VecPoly3> {
+        let mut polys = ArrayTrait::new();
+        loop {
+            match scalars.pop_front() {
+                Option::Some(scalar) => {
+                    polys.append(VecPoly3Trait::single_scalar_poly(*scalar));
+                },
+                Option::None(_) => {
+                    break;
+                },
+            };
+        };
+        polys
+    }
+}
+
+
+/// Represents a vector used in the verification MSM,
+/// specifying the next index to use, and, for the s vectors, the max index to use
+#[derive(Drop, Serde, PartialEq, Copy)]
+enum VecElem {
+    /// The flattened vector of left weights
+    w_L_flat: usize,
+    /// The flattened vector of right weights
+    w_R_flat: usize,
+    /// The flattened vector of output weights
+    w_O_flat: usize,
+    /// The flattened vector of witness weights
+    w_V_flat: usize,
+    /// The vector of G generator coefficients in the IPA proof
+    s: (usize, usize),
+    /// The vector of H generator coefficients in the IPA proof
+    s_inv: (usize, usize),
+}
+
+/// Represents the remaining powers of a challenge scalar needed for verification.
+/// We always do consecutive exponentiation, so it's sufficient to just store the
+/// base, the last power computed and the number of exponentiations left to do.
+#[derive(Drop, Serde, PartialEq, Copy)]
+struct RemainingScalarPowers {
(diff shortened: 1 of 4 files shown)


src/verifier/types.cairo
```diff
@@ -0,0 +1,202 @@
...
+    /// Tracks the powers of challenge scalar z left to compute
+    z_powers_rem: RemainingScalarPowers,
+    /// Tracks the G generators left to sample from the hash chain
+    G_rem: RemainingGenerators,
+    /// Tracks the H generators left to sample from the hash chain
+    H_rem: RemainingGenerators,
+    /// The proof-specific commitments remaining to be used for verification
+    commitments_rem: Array<EcPoint>,
+    /// The accumulated result of the verification MSM
+    msm_result: Option<EcPoint>,
```
joeykraut, line 29: Why an `Option` here? Can we just use the additive identity?

src/verifier/types.cairo
```diff
@@ -0,0 +1,202 @@
...
+    // The final verdict of the verification
+    verified: Option<bool>,
```
joeykraut: Similar thing here, can this just be `false` until the `VerificationJob` has been stepped through? I feel that `Option` clutters the representation

src/verifier/types.cairo
```diff
@@ -0,0 +1,202 @@
...
+    /// The number of multiplication gates in the circuit
+    n: usize,
+    /// The number of multiplication gates in the circuit, padded to the next power of 2
+    n_plus: usize,
+    /// log_2(n_plus)
+    k: usize,
+    /// The number of linear constraints in the circuit
+    q: usize,
+    /// The size of the witness
+    m: usize,
```
joeykraut, line 183: I feel we should have more meaningful names here; e.g. `num_multiplications`, `num_multiplications_padded`, etc. I'm fine leaving this for now if it makes implementation significantly easier though; i.e. if this is the same notation used in Dalek

src/verifier/utils.cairo
```diff
@@ -0,0 +1,84 @@
...
+use renegade_contracts::utils::{math::dot_product};
+
+use super::types::{VerificationJob, SparseWeightMatrixSpan, SparseWeightVecSpan};
+
+// --------
+// | MATH |
+// --------
+
+fn flatten_sparse_weight_matrix(
+    matrix: SparseWeightMatrixSpan, z_powers_to_q: Span<felt252>, width: usize,
```
joeykraut: nit: maybe just `z_powers` here? I feel the full name is more confusing

src/verifier/utils.cairo
```diff
@@ -0,0 +1,84 @@
...
+
+use renegade_contracts::utils::{math::dot_product};
+
+use super::types::{VerificationJob, SparseWeightMatrixSpan, SparseWeightVecSpan};
+
+// --------
+// | MATH |
+// --------
+
+fn flatten_sparse_weight_matrix(
```
joeykraut, line 17: Add a doc comment here explaining what this computes mathematically, is it $z^TW$, $Wz$, etc?

src/verifier/utils.cairo
```diff
@@ -0,0 +1,84 @@
...
+            let weight = get_weight_or_zero(ref row, col_index);
+            column.append(weight);
+
+            row_index += 1;
+        };
+
+        // Dot product z_powers_to_q with the column
+        flattened.append(dot_product(z_powers_to_q, column.span()));
+        col_index += 1;
+    };
```
joeykraut: Do we really need to materialize the whole matrix in this method (when most of it is zero)? Seems wasteful. Like rows will often have less than 1% density. Wouldn't a better approach be to index into the `z_powers` array directly and build a dot product out of only the terms that are non-zero?

src/verifier/utils.cairo
```diff
@@ -0,0 +1,84 @@
...
+            // Unwrapping is safe here since we know sparse_vec.len() > 0
+            let (_, current_weight) = *sparse_vec.pop_front().unwrap();
+            return current_weight;
+        }
+    }
+
+    // If the sparse vec is empty, or the index of the current tuple
+    // is greater then expected the weight must be zero
+    0
+}
```
joeykraut: this method won't really work correctly if not used the exact way you are currently; which means it's a tight coupling. For example; if `sparse_vec = [(1, x), (2, y)]` and I call `get_weight_or_zero(ref sparse_vec, 2)` I'll get `0`, not `y`. 

I'd either refactor this to work properly for all inputs, fold this into its caller if we want to maintain the coupling, or take the suggestion I have above about doing sparse multiplication.

joeykraut (changes requested):
A few things I'd like to address before approving, but the abstractions are looking strong. It's a bit hard to gauge where this is going in the broader scope of the verifier, so I may miss some things. Let's discuss in person if you'd like to chat about the comments here

src/verifier/types.cairo
akirillo: Perhaps a bit paranoid, but I'm worried about code paths that lead to an early false verification, since the final check is `msm_result = ec_point_zero()`

src/verifier/types.cairo
akirillo: Yeah, here should be fine to just use false tbh, will do

src/verifier/types.cairo
akirillo: yeah, this is the notation used in Dalek, which I'm quite used to. I'm actually unsure if we should move off of this notation, because I imagine that auditors / interested readers will/should be inspecting the code side-by-side with the Dalek writeup (at least, that's what I'll probably suggest in the docs)

src/verifier/utils.cairo
akirillo: yeah, I just wanted to write something that was easier to reason about at first, but in my branch that implements the processing of the MSM, what you describe is what I do. I can bring over that implementation here

src/verifier/utils.cairo
akirillo: ah yeah, i see what you mean - it's not as flexible as the name lets on to be. as per the comment above though, I'll just move over the implementation of sparse multiplication

src/verifier/utils.cairo
```diff
@@ -0,0 +1,93 @@
...
+        };
+    };
+
+    res
+}
+
+/// Given a sparse-reduced circuit weight matrix (W_{L, R, O, V}), this extracts
+/// the column at `desired_col_index` as a sparse-reduced vector
+fn get_sparse_weight_column(
+    mut matrix: SparseWeightMatrixSpan, desired_col_index: usize
```
joeykraut: nit: just `col_index` or `col`

src/verifier/utils.cairo
```diff
@@ -0,0 +1,93 @@
...
+
+use super::types::{VerificationJob, SparseWeightMatrixSpan, SparseWeightVec, SparseWeightVecSpan};
+
+// --------
+// | MATH |
+// --------
+
+/// Given a sparse-reduced circuit weight matrix (W_{L, R, O, V}) with (actual)
+/// width `width`, "flattens" the matrix into a `width`-length vector by computing
+/// [z, z^2, ..., z^width] * W_{L, R, O, V} (vector-matrix multiplication)
```
joeykraut: This should be `[z, z^2, ..., z^height]` not `z^width`

src/verifier/utils.cairo
```diff
@@ -0,0 +1,93 @@
...
+    let mut column = ArrayTrait::new();
+    let mut row_index = 0;
+    loop {
+        if row_index == matrix.len() {
+            break;
+        };
+
+        let mut row = *matrix.at(row_index);
+        loop {
+            match row.pop_front() {
```
joeykraut: Won't `pop_front` mutate the row? Then if it's not the correct `col_index` we need to add it back to the vector in the first position. Feels like we should peek the element instead. Or just use the direct method I suggest above

src/verifier/utils.cairo
```diff
@@ -0,0 +1,93 @@
...
+
+    // Vector-matrix multiplication => loop over columns first, then rows
+    let mut col_index: usize = 0;
+    loop {
+        if col_index == width {
+            break;
+        };
+
+        // Get column as a SparseWeightVec
+        let mut column = get_sparse_weight_column(matrix, col_index).span();
```
joeykraut: We are still doing a lot more work than is necessary here, i.e. looping over the entire matrix, this defeats the point of having a sparse representation (computationally).

I'd suggest an approach that doesn't directly loop like this, but instead only accesses elements it needs. Here's some psuedocode in python-ish assuming a row-major repr as we have
```python3
# Computes zW for vector z (1 x n) and matrix W (n x m), returinng a 1 x m vector
def lhs_vector_mat_mul(vec: List[int], sparse_mat: List[List[int]]) -> List[int]:
    res = [0 for _ in len(sparse_mat[0])] # width-length zero'd buffer
    for (row_ind, row) in enumerate(sparse_mat):
         vec_term = vec[row_ind]
         for (col_ind, val) in row:
              res[col_ind] += vec_term * val
    res
```

This probably means removing the helper below so that we don't have to loop in the way we currently are.

joeykraut (changes requested):
A few more things I'd like to cover before approving

src/verifier/utils.cairo
akirillo: LOL yes 🤦

src/verifier/utils.cairo
akirillo: Ah, I can leave a comment explaining this: the type we're iterating over here is a `Span`, which is a wrapper around an `Array` snapshot that still exposes the array API.

Using `pop_front` on a `Span` doesn't mutate the underlying `Array`, just the `Span`/snapshot

joeykraut: But we do this in a loop, so the *same* underlying span is used between iterations

src/verifier/utils.cairo
```diff
@@ -0,0 +1,90 @@
...
+            if entry_index == row.len() {
+                break;
+            };
+
+            let (col_index, weight) = *row.at(entry_index);
+            let col_index_felt = col_index.into();
+            // Default value for an unset key is 0
+            let mut value = flattened_dict.get(col_index_felt);
+            // z vector starts at z^1, i.e. is [z, z^2, ..., z^q]
+            let z_i = binary_exp(z, row_index + 1);
```
joeykraut, line 45: we can incrementally compute `z_i` as we loop over rows, no need to do `binary_exp` each time

src/verifier/utils.cairo
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    flattened_vec
+}
+
+/// Given a sparse-reduced column vector `col`, "flattens" the vector into a
+/// single scalar by computing sum(z^i * col[i]) for all indices i with non-zero
+/// weights in the column. This is effectively a dot product [z, z^2, ..., z^len(col)] * col,
+/// but omitting multiplications by zero.
+fn flatten_column(mut col: SparseWeightVecSpan, z: felt252) -> felt252 {
```
joeykraut, line 73: is this method still used? Also we can compute the `z_i` incrementally here as well if so

joeykraut (approved):
two smaller things, lgtm

src/verifier/utils.cairo
akirillo: yes, it will be used in computing $w_c = \vec{z} \cdot \vec{c}$.
I specifically opted against incremental computation here because I'm betting on a rather sparse $\vec{c}$

src/verifier/utils.cairo
akirillo: Hmm, I opted against this as it's possible to have empty rows - I'm making a bet on sparseness here

PR #21, opened 2023-06-17
renegade_contracts: verifier: impl initialize fn

akirillo:
This PR stubs out the contract structure using Cairo `v2.0.0-rc2` syntax, and implements the `initialize` function on the verifier contract.

For now, we assume that all types can be stored in a single storage address (which can hold 256 slots / felts).

CI is red, but the `test_initializer_storage_serde` test passes, asserting that circuit params are written to contract storage correctly.

### DIFF
--- a/src/verifier.cairo
+++ b/src/verifier.cairo
@@ -1,2 +1,136 @@
 mod types;
 mod utils;
+
+// -------------
+// | INTERFACE |
+// -------------
+// TODO: Move to separate file / module?
+
+use types::CircuitParams;
+
+#[starknet::interface]
+trait IVerifier<TContractState> {
+    fn initialize(ref self: TContractState, circuit_params: CircuitParams);
+    fn get_circuit_params(self: @TContractState) -> CircuitParams;
+}
+
+#[starknet::contract]
+mod Verifier {
+    use traits::Into;
+    use ec::EcPoint;
+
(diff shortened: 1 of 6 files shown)


src/verifier.cairo
```diff
@@ -1,2 +1,136 @@
...
+    // ----------------------------
+    // | INTERFACE IMPLEMENTATION |
+    // ----------------------------
+
+    #[external(v0)]
+    impl IVerifierImpl of super::IVerifier<ContractState> {
+        /// Initializes the verifier for the given public parameters
+        /// Parameters:
+        /// - `circuit_params`: The public parameters of the circuit
+        fn initialize(ref self: ContractState, circuit_params: CircuitParams) {
```
joeykraut, line 88: Just calling this out so we're absolutely clear, this method will be called by the main contract's `initialize` method which has appropriate access control + double initialize safeguards?

joeykraut (approved):
looking great, super clean, just had a question for my own understanding :)

src/verifier.cairo
akirillo: yes, exactly - will refactor to logically separate library/impl & actual contract w/ external fns, access & safety controls once core logic is set down and once the extensibility design laid out [here](https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794#extensibility-and-components-11) has crystallized / been implemented in the compiler

PR #22, opened 2023-06-17
renegade_contracts: verifier: impl queue_verification_job fn

akirillo:
This PR introduces the `queue_verification_job` function, which preps a verification job so that the underlying MSM can later be processed.

For now, we make the optimistic assumption that all of the required scalar/felt computation fits within a transaction - this is relevant in the context of calculating delta, a scalar term in the MSM which requires flattening 2 weight matrices and the constants vector - this takes a lot of work (asymptotically).

The `test_queue_verification_job` test passes, for which I evaluated the structure & scalar computation for the given circuit by hand, this is what's hardcoded in the test case.

### DIFF
--- a/src/verifier.cairo
+++ b/src/verifier.cairo
@@ -6,22 +6,39 @@ mod utils;
 // -------------
 // TODO: Move to separate file / module?
 
-use types::CircuitParams;
+use types::{CircuitParams, Proof, VerificationJob};
 
 #[starknet::interface]
 trait IVerifier<TContractState> {
     fn initialize(ref self: TContractState, circuit_params: CircuitParams);
+    fn queue_verification_job(ref self: TContractState, proof: Proof, verification_job_id: felt252);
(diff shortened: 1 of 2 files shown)


src/verifier.cairo
```diff
@@ -110,6 +127,217 @@ mod Verifier {
             self.c.write(StorageAccessSerdeWrapper { inner: circuit_params.c });
         }
 
+        /// Enqueues a verification job for the given proof, squeezing out challenge scalars
+        /// as necessary.
+        /// Parameters:
+        /// - `proof`: The proof to verify
+        /// - `verification_job_id`: The ID of the verification job
+        fn queue_verification_job(
```
joeykraut, line 135: this all largely looks good, copying here what we talked about IRL for posterity: It'd be nice to have little comments as mental hooks that refer to what each component of the MSM does at each step in the setup. These can refer to the docs too if we like

joeykraut (approved):
lgtm

PR #23, opened 2023-06-23
renegade_contracts: utils: unit testing & doc comments for utils

akirillo:
This PR adds basic unit tests & doc comments for relevant utils.
Test suite is not yet expansive, looking to scope out more time later on to sit down and rack my brain for good unit, integration, fuzz, etc tests once the verifier is closer to being crystallized


joeykraut (approved):
lgtm

PR #24, opened 2023-06-23
renegade_contracts: verifier: pack verifier utils into sep. files / traits

akirillo:
This PR moves the `calc_delta` and `squeeze_challenge_scalars` helper functions into the `verifier/utils.cairo` folder, and moves some sparse weight matrix / vector utilities (i.e. flattening) into a trait, allowing relevant methods to be called directly on the types.

### DIFF
--- a/src/testing/tests/verifier_tests.cairo
+++ b/src/testing/tests/verifier_tests.cairo
@@ -168,11 +168,14 @@ fn test_calc_delta_basic() {
         .append(3350465545061232605275298873236176023725099273455182129604714866792474093038);
 
     let delta = calc_delta(n, y_inv_powers_to_n.span(), z, @W_L, @W_R);
-    delta.print();
-    assert(
-        delta == 1206167596222043737899107594365023368541035738443865566657697352045290674603,
-        'wrong delta'
-    );
+
+    // delta = <y^{n+}[0:n] * w_R_flat, w_L_flat>
+    // The expected result was calculated by hand using the powers of y^{-1} above
+    // and the result of flattening the matrices W_L and W_R above using z = 2
+    let expected_delta =
+        1206167596222043737899107594365023368541035738443865566657697352045290674603;
+
+    assert(delta == expected_delta, 'wrong delta');
 }
 
 // ------------------
@@ -525,6 +528,9 @@ fn get_expected_verification_job() -> VerificationJob {
     }
 }
 
+// The `scalar` fields in these terms were calculated by hand using the following values,
+// given from the placeholder `squeeze_challenge_scalars` implementation & the `get_dummy_proof` helper:
+// y = 2, z = 3, x = 4, w = 5, u = [6], r = 8, t_hat = 9, t_blind = 10, e_blind = 11, a = 12, b = 13
 fn get_expected_scalars_rem() -> Array<VecPoly3> {
     let mut scalars_rem = ArrayTrait::new();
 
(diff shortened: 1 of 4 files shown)


src/testing/tests/verifier_tests.cairo
```diff
@@ -1,22 +1,184 @@
...
+        .append(2814391057851435388431251053518387859929083389702352988867960488105678238152);
+    y_inv_powers_to_n
+        .append(3350465545061232605275298873236176023725099273455182129604714866792474093038);
+
+    let delta = calc_delta(n, y_inv_powers_to_n.span(), z, @W_L, @W_R);
+    delta.print();
+    assert(
+        delta == 1206167596222043737899107594365023368541035738443865566657697352045290674603,
+        'wrong delta'
+    );
```
joeykraut: Did we calculate these by hand or pull them from some implementation? Would be good to call out some of these magic numbers in comments

src/verifier/types.cairo
```diff
@@ -168,6 +177,146 @@ type SparseWeightMatrix = Array<SparseWeightVec>;
...
+                if entry_index == row.len() {
+                    break;
+                };
+
+                let (col_index, weight) = *row.at(entry_index);
+                let col_index_felt = col_index.into();
+                // Default value for an unset key is 0
+                let mut value = flattened_dict.get(col_index_felt);
+                // z vector starts at z^1, i.e. is [z, z^2, ..., z^q]
+                let z_i = binary_exp(z, row_index + 1);
```
joeykraut: This can be moved to the outer loop

src/verifier/types.cairo
```diff
@@ -168,6 +177,146 @@ type SparseWeightMatrix = Array<SparseWeightVec>;
...
+            if row_index == matrix.len() {
+                break;
+            };
+
+            let mut row = *matrix.at(row_index);
+            let mut entry_index = 0;
+            loop {
+                if entry_index == row.len() {
+                    break;
+                };
```
joeykraut: We can break earlier if `entry_index > col_index` right? The rows should be sorted by index unless I'm misunderstanding

joeykraut: If you make this change, maybe also call out this sorting assumption in a comment

src/verifier/types.cairo
```diff
@@ -168,6 +177,146 @@ type SparseWeightMatrix = Array<SparseWeightVec>;
...
+    fn assert_width(self: @SparseWeightMatrix, width: usize) {
+        let matrix: SparseWeightMatrixSpan = self.deep_span();
+        let mut row_index = 0;
+        loop {
+            if row_index == matrix.len() {
+                break;
+            };
+
+            let row = *matrix.at(row_index);
+            assert(row.len() <= width, 'row too wide');
```
joeykraut: The length of the sparse representation of a row and the width of the underlying matrix are different here no? My assumption is that this method computes the "rightmost column appearing in any row" in which case the psuedocode would be more like
```python
max([row[-1].index for row in self])
```

If I'm just misunderstanding the point of this method would be good to disambiguate in the doc comment

joeykraut (approved):
some smaller optimizations and suggestions, would prefer we at least discuss in comments before merging but overall looks good enough for approval

src/testing/tests/verifier_tests.cairo
akirillo: calculated all of them by hand, with the equations in the comments. will call this out

src/verifier/types.cairo
akirillo: yes, you're correct on sorting assumption - will call this out

src/verifier/types.cairo
akirillo: mm, the point of this method was just to assert that the row has at most `width` entries, though what you call out is something else worth asserting. I'll add that as another check and update the doc comment.

PR #25, opened 2023-06-23
renegade_contracts: verifier: refactored VerificationJob types

akirillo:
This PR refactors some of the types used in the verifier.

Primarily, we reduce the logic of "exhausting" a scalar polynomial to only be when the vector term it relies on (if any) is exhausted, i.e. we used the last element of the vector. Previously, this logic also included "if we use the last power of $y^{-1}$ this polynomial requires". The latter clause is unnecessary as it is covered by the vector term being used up, that's when we stop exponentiating $y^{-1}$. 

We also move the current index of a polynomial term's vector element out of the `VecPoly3Term` and into the top-level `VerificationJob`, i.e. for all vectors $\vec{w}_{\{L,R,O,V\}}, \vec{s}, \vec{s}^{-1}, \vec{u}^2, \vec{u}^{-2}$, we track the index of the next element to evaluate in a single struct stored on the `VerificationJob`.
Previously, we stored in this inside the `VecPoly3Term` itself, but this was an oversight on my part - in Cairo, it's impossible to update an `Array` element in-place, you have to deconstruct and reconstruct the entire `Array`. So updating these indices inside the `VecPoly3 = Array<VecPoly3Term>` would've been impossible. This also reduces storage costs somewhat.

Additionally, we simply store $z$ in the `VerificationJob`, as we don't have a chance to incrementally compute powers across verification steps as we do with $y^{-1}$, and we store the $\vec{u}$ challenge scalars on the `VerificationJob` as well as they are needed for computing the $\vec{s}, \vec{s}^{-1}$ values.

Finally, I change the circuit used in the testing code such that $n \neq n^+$, i.e. the actual number of multiplication gates is less than the padded number. This is to cover more code paths, b/c when $n = n^+$, we omit some scalar polynomials from the final MSM.

All tests pass when run with `scarb test`.

Happy to chat about this in person if it wasn't conveyed well in the PR description.

### DIFF
--- a/src/testing/tests/verifier_tests.cairo
+++ b/src/testing/tests/verifier_tests.cairo
@@ -12,8 +12,8 @@ use renegade_contracts::{
         Verifier,
         types::{
             SparseWeightMatrix, SparseWeightMatrixTrait, SparseWeightVec, SparseWeightVecTrait,
-            CircuitParams, Proof, VerificationJob, RemainingScalarPowers, RemainingGenerators,
-            VecPoly3Term, VecPoly3, VecElem
+            CircuitParams, Proof, VerificationJob, VerificationJobTrait, RemainingGenerators,
+            RemainingGeneratorsTrait, VecPoly3Term, VecPoly3, VecSubterm, VecIndices
         },
         utils::{get_s_elem, calc_delta},
     },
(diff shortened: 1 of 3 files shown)


src/testing/tests/verifier_tests.cairo
```diff
@@ -184,12 +184,12 @@ fn test_calc_delta_basic() {
...
-// n = 2
-// n_plus = 2
-// k = log2(n_plus) = 1
+// 3 multiplication gates, and 2 linear constraints. In total, it is parameterized as follows:
+// n = 3
+// n_plus = 4
+// k = log2(n_plus) = 2
 // m = 4
-// q = 6 (The total number of linear constraints is 6 because there are 2 multiplication gates, and 2 linear constraints per multiplication gate)
+// q = 8 (The total number of linear constraints is 6 because there are 3 multiplication gates, and 2 linear constraints per multiplication gate)
```
joeykraut: nit: should be "The total number of linear constraints is 8..."

joeykraut (approved):
nice one

PR #26, opened 2023-06-23
renegade_contracts: verifier: impl step verification

akirillo:
This PR implements the step verification logic.

The basic pattern is to evaluate the next scalar polynomial, update necessary state (power of $y^{-1}$, index w/in vectors, $G$ / $H$ generators), and pop the polynomial if its exhausted.

Then compute / pop the next EC point, do an `EC OP`, and accumulate that into `msm_result` until we run out of gas, or the verification is complete (i.e., there are no more scalar polynomials remaining).

The way we determine whether or not enough gas remains is pretty rudimentary, for now. The most expensive step of the MSM's cost is (asymptotically) dominated by flattening matrices, which has cost $O(qn)$ where $q$ is the number of linear constraints and $n$ is the number of mul gates

### DIFF
--- a/src/verifier.cairo
+++ b/src/verifier.cairo
@@ -12,6 +12,7 @@ use types::{CircuitParams, Proof, VerificationJob};
 trait IVerifier<TContractState> {
     fn initialize(ref self: TContractState, circuit_params: CircuitParams);
     fn queue_verification_job(ref self: TContractState, proof: Proof, verification_job_id: felt252);
+    fn step_verification(ref self: TContractState, verification_job_id: felt252);
     fn get_circuit_params(self: @TContractState) -> CircuitParams;
     fn get_verification_job(self: @TContractState, verification_job_id: felt252) -> VerificationJob;
 }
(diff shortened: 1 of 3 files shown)


src/verifier.cairo
```diff
@@ -369,6 +372,32 @@ mod Verifier {
                 .write(verification_job_id, StorageAccessSerdeWrapper { inner: verification_job });
         }
 
+        fn step_verification(ref self: ContractState, verification_job_id: felt252) {
```
joeykraut, line 375: What're your thoughts on adding access control to these methods? I know this is a library and we'd likely want to handle this at a higher module; but I'm curious what your plan is when we integrate this around access control. Do we need it?

joeykraut (approved):
clean abstractions, nice

src/verifier.cairo
akirillo: yeah, putting off access controls until we integrate w/ the darkpool contract, as there's more stuff to do on that front first (e.g. statement injection).

with that said, one can think of some interesting access control patterns on this function in particular... but I'm inclined to think we actually need _none_. If anyone wants to drive my verification job forward, go ahead lol

PR #27, opened 2023-06-24
docs: init verifier docs

akirillo:
This PR introduces some high-level documentation of the verifier's design, basically paring down what we've already documented internally.

The parts of the design I've chosen to focus on are:
- The loop-till-out-of-gas concept
- The split between `queue_verification_job` & `step_verification`
- Representing MSM scalars as vector polynomials 
- Using sparse-reduced representation of weight vectors

Let me know if there are any other topics I should mention here, or if you'd like to see more detail on any of them. This is just a starting point for now

### DIFF
--- a/docs/verifier.md
+++ b/docs/verifier.md
@@ -1,0 +1,194 @@
+# Verifier
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


akirillo:
oh that's incredibly cool and based how github parses latex differently from vscode

i'll figure this out & fix the formatting

PR #28, opened 2023-07-12
renegade_contracts: transcript: basic Fiat-Shamir transformation

akirillo:
This PR introduces a very basic Fiat-Shamir transformation based on a Keccak hash chain of the transcript, along with a counter that gets hashed into the transcript every time a value is squeezed out (to allow for consecutive squeezes).

The bulk of this diff is me moving around some verifier code into helpers that let me continue to unit test the verifier with dummy challenge scalars (now that we are actually drawing real challenges from the transcript).

Something I want to follow up on is the extra bit-space needed to get a uniform sampling of challenge scalars. Right now, we simply reduce the `u256` returned from a Keccak hash into a field element, which, I would think, should still bear a uniform sampling of field elements (since `u256` is bigger than a `felt252`), but @joeykraut I know in the relayer we reduce 48-byte challenges into field elements, so this something I still have to implement.


src/testing/tests/verifier_tests.cairo
```diff
@@ -527,6 +497,83 @@ fn get_dummy_circuit_pc_gens() -> (EcPoint, EcPoint) {
...
+    let circuit_params = get_dummy_circuit_params();
+
+    'initializing...'.print();
+    verifier.initialize(circuit_params.clone());
+
+    circuit_params
+}
+
+// Mostly copied from the contract, but uses dummy proof & challenge scalars,
+// and returns verification job instead of writing to contract state.
```
joeykraut: nit: just make doc comments describe the function, not as they relate to other doc comments, this makes it hard to read in isolation

joeykraut:
> This PR introduces a very basic Fiat-Shamir transformation based on a Keccak hash chain of the transcript, along with a counter that gets hashed into the transcript every time a value is squeezed out (to allow for consecutive squeezes).
> 
> The bulk of this diff is me moving around some verifier code into helpers that let me continue to unit test the verifier with dummy challenge scalars (now that we are actually drawing real challenges from the transcript).
> 
> Something I want to follow up on is the extra bit-space needed to get a uniform sampling of challenge scalars. Right now, we simply reduce the `u256` returned from a Keccak hash into a field element, which, I would think, should still bear a uniform sampling of field elements (since `u256` is bigger than a `felt252`), but @joeykraut I know in the relayer we reduce 48-byte challenges into field elements, so this something I still have to implement.

It will not give a uniform sampling of the field. In fact, no amount of bits will, it's just about an amount that gives a distribution that is computationally indistinguishable from uniform. 31 extra bits is *not* computationally indistinguishable, 128 bits (16 bytes, so 32 bytes + 16 bytes = 48 that we squeeze) *is* computationally indistinguishable.

src/utils/math.cairo
```diff
@@ -53,3 +56,7 @@ fn binary_exp(base: felt252, exp: usize) -> felt252 {
         base * binary_exp(base * base, (exp - 1) / 2)
     }
 }
+
+fn reduce_to_felt(val: u256) -> felt252 {
```
joeykraut: Add a todo to update this to the correct version we use prover side

joeykraut (approved):
Looking clean

src/testing/tests/verifier_tests.cairo
akirillo: This is a good point - updated the comment to describe the function itself, but also left in this note as I think it's helpful context

src/utils/math.cairo
```diff
@@ -53,3 +57,27 @@ fn binary_exp(base: felt252, exp: usize) -> felt252 {
...
+    data.append(hash);
+    let high_u256 = keccak_u256s_be_inputs(data.span());
+
+    let high_u128 = high_u256.low;
+
+    let low_felt = (hash % STARK_FIELD_PRIME).try_into().unwrap(); // low_u256 % p
+    let high_felt = (high_u128.into() % STARK_FIELD_PRIME).try_into().unwrap(); // high_u128 % p
+
+    low_felt + (high_felt * SHIFT_128)
+}
```
akirillo: Just realized that my rebasing will probably make it difficult to view changes since your last review, this is really the only relevant new change

joeykraut (approved):
Just reviewed the changes to `hash_to_felt`, lgtm

PR #29, opened 2023-07-13
renegade_contracts: verifier: add verification events

akirillo:
This PR introduces a minimal set of events for verifier initialization, verification job queueing, and verification job completion, as well as basic testing for event emission.


joeykraut (approved):
looks good

PR #30, opened 2023-07-19
renegade_contracts: verifier: scalar: `Scalar` type

akirillo:
This PR replaces usage of STARK base field elements (`felt252`) with STARK scalar field elements using a new `Scalar` type & appropriate operator overloading.

A large portion of this diff is me updating the unit tests to reflect the proper calculations in the scalar field - this was (still) done by hand and the tests pass. No structural changes were made to the tests so you can ignore them during review.

Additionally, I branched this off of my WIP branch for updating the transcript interface, since I made a lot of changes there that it would be easier to build this on top of, and I need thwse changes to test that branch for transcript consistency with the relayer. PR for that coming soon.


src/verifier/scalar.cairo
```diff
@@ -0,0 +1,170 @@
...
+    fn mul_eq(ref self: Scalar, other: Scalar) {
+        self = Mul::mul(self, other);
+    }
+}
+
+// ------------
+// | DIVISION |
+// ------------
+
+impl ScalarDiv of Div<Scalar> {
```
joeykraut, line 89: nit: Can we document the behavior of division `div(a, b)` when $b \nmid a$? It's not immediately clear, is this truncated, rounded, etc?

joeykraut (approved):
One small clarification ask, otherwise looks very good. I like that wrapping with the `Scalar` type cleans up many of the arithmetic interfaces, e.g. `neg` and `Scalar::inverse`

PR #31, opened 2023-07-20
renegade_contracts: verifier, transcript: consistency with prover-side implementation

akirillo:
This PR brings the verifier code into a state that is consistent with the prover-side implementation, i.e. this actually verifies proofs for the example circuit! 🎉 

Some of my prior implementation was actually incorrect (e.g. the calculation of $\vec{s}$), so not all tests pass - but the next step for me is doing some _light_ revamping of the testing stack, so I didn't bother redoing all the math by hand for this PR.

The important test that _does_ pass is `test_full_verification_ex_proof` - I'll let you guess what it does ;)

Apologies for some git grossness here - I merged #30 onto this branch which really beefs up the diff, then put a commit on top. The actual new changes to review are changes from commits 1-3, then commit 5.


src/testing/tests/verifier_tests.cairo
```diff
@@ -189,10 +214,102 @@ fn test_squeeze_challenge_scalars_basic() {
...
 
-    assert(challenge_scalars.len() == 5, 'wrong # of challenge scalars');
+    assert(challenge_scalars.len() == 6, 'wrong # of challenge scalars');
     assert(u.len() == k, 'wrong # of u challenge scalars');
 }
 
+#[test]
+#[available_gas(100000000)]
+fn test_transcript_basic() {
+    // Obtained by processing the below transcript in the prover-side implementation
```
joeykraut, line 224: nit: code link would be helpful here

src/testing/tests/verifier_tests.cairo
```diff
@@ -323,6 +440,27 @@ fn test_step_verification_basic() {
     assert(verification_job.msm_result.unwrap() == get_expected_msm_result(), 'wrong msm_result');
 }
 
+#[test]
+#[available_gas(1000000000)] // 10x
+fn test_full_verification_ex_proof() {
```
joeykraut, line 446: out of completeness it'd be good to have a test for a modified proof. I.e. take the `get_example_proof()` output and modify it to ensure it doesn't verify

src/transcript.cairo
```diff
@@ -14,28 +20,114 @@ struct Transcript {
...
+    /// Append a `point` with the given `label`.
+    fn append_point(ref self: Transcript, label: u256, point: EcPoint) {
+        let mut message = ArrayTrait::new();
+        match ec_point_is_zero(point) {
+            IsZeroResult::Zero(()) => {
+                // x
+                message.append(0.into());
+                // y
+                message.append(0.into());
+            },
```
joeykraut, line 98: The point at infinity should result in an error if added to the transcript, we do not add the affine coordinates directly.

joeykraut (approved):
few small things, lgtm

PR #32, opened 2023-07-21
renegade_contracts: verifier: separated witness from proof

akirillo:
This PR separates the witness commitments ($\vec{V}$) from the proof struct - this both more closely matches the Bulletproofs specification and will make witness injection (e.g. adding merkle root to witness) more ergonomic down the line

### DIFF
--- a/src/testing/tests/verifier_tests.cairo
+++ b/src/testing/tests/verifier_tests.cairo
@@ -211,8 +211,11 @@ fn test_calc_delta_basic() {
 #[available_gas(100000000)]
 fn test_squeeze_challenge_scalars_basic() {
     let proof = get_dummy_proof();
+    let witness_commitments = get_dummy_witness_commitments();
     let (_, n_plus, k, _, m) = get_dummy_circuit_size_params();
-    let (challenge_scalars, u) = squeeze_challenge_scalars(@proof, m, n_plus);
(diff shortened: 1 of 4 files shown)


src/testing/tests/verifier_tests.cairo
```diff
@@ -362,6 +370,7 @@ fn test_initializer_storage_serde() {
 }
 
 #[test]
+#[ignore]
```
joeykraut, line 373: For my own understanding, what does this macro do?

joeykraut (approved):

src/testing/tests/verifier_tests.cairo
akirillo: the test runner won't run this test unless explicitly specified using the `--include-ignored` flag

the tests I've marked as `#[ignore]` are either 1) expected to fail because I didn't redo the math over the scalar field for them, or 2) only print debug outputs and don't assert anything about their logic

I've left those tests in the repo for now since I want to soon refactor the testing stack and may want to refer the tests to reimplement them, but at the same time I'm happy to remove them and dig them up from git history later...

akirillo: thought about it, and ended up doing this in #33

PR #35, opened 2023-07-28
renegade_contracts: merkle: upgrade to Cairo v2.0.1

akirillo:
This PR ugrades the Merkle tree contract to target v2.0.1 of the Cairo compiler (commit message is wrong, we're building using 2.0.1).

Ignore the `test_devnet` action for this PR, I re-enabled it for the test-stack refactor PRs


src/merkle.cairo
```diff
@@ -1,32 +1,287 @@
-mod interface;
-mod library;
+#[starknet::interface]
+trait IMerkle<TContractState> {
+    fn initializer(ref self: TContractState, height: u8);
+    fn get_root(self: @TContractState) -> felt252;
+    fn root_in_history(self: @TContractState, root: felt252) -> bool;
```
joeykraut: Should the root be the `Scalar` type we have defined elsewhere? The entirety of Merkle tree should operate over the same field as the proof system.

src/merkle.cairo
```diff
@@ -1,32 +1,287 @@
...
 
-    #[external]
-    fn initializer(height: u8) {
-        MerkleLib::initializer(height);
+    /// The value of an empty leaf in the Merkle tree:
+    /// 306932273398430716639340090025251549301604242969558673011416862133942957551
+    /// This value is computed as the keccak256 hash of the string 'renegade'
+    /// taken modulo the Cairo field's prime modulus (see `BASE_FIELD_ORDER` in src/utils/constants.cairo)
+    const EMPTY_LEAF_VAL: felt252 =
+        306932273398430716639340090025251549301604242969558673011416862133942957551;
```
joeykraut: Yeah again, we need to take this modulo the scalar field since we're proving over the scalar field

src/merkle.cairo
```diff
@@ -1,32 +1,287 @@
...
+    // -----------
+
+    #[storage]
+    struct Storage {
+        /// The height of the Merkle tree stored by this contract
+        height: u8,
+        /// Capacity of the Merkle tree, cached in contract storage
+        capacity: u128,
+        /// The next index to insert a node at
+        next_index: u128,
```
joeykraut: Probably worth documenting that the `next_index` field and the `root_history` will always be off by one. i.e. we initialize `root_history[initial_root] = 1`, but initialize `next_index = 1`; meaning the original root is interpreted as being at insertion index `0`. 

While we don't currently rely on any relationship between these values, I could see a world in which we do, and this would be nice to call out

src/merkle.cairo
```diff
@@ -1,32 +1,287 @@
...
+            .emit(
+                Event::MerkleInternalNodeChanged(
+                    MerkleInternalNodeChanged { height, index: insert_index, new_value: value }
+                )
+            );
+
+        // Base case
+        if height == 0 {
+            return value;
+        }
```
joeykraut: Switch the order of these two? no need to emit an event at the root, this is handled by a separate event

src/testing/tests/merkle_tests.cairo
```diff
@@ -0,0 +1,55 @@
+use traits::Into;
```
joeykraut, line 1: Do we have any tests against a known working implementation (e.g. Arkworks)?

joeykraut (commented):
Looks good, want to discuss which field this is over before approving

src/merkle.cairo
akirillo: ah, was unsure about this so I left it as a felt for now, I'll move it over to `Scalar`

src/testing/tests/merkle_tests.cairo
akirillo: heh, that's coming in the other PRs ser

src/merkle.cairo
```diff
@@ -1,32 +1,292 @@
...
+
+        // The next value in the sibling pathway is the current hash, when the first value
+        // is inserted into the Merkle tree, it will be hashed against the same values used
+        // in this recursion
+        self.sibling_path.write(height, current_leaf);
+
+        // Hash the current leaf with itself and recurse
+        // The `.into()` call here reduces the `felt252` output of the Pedersen hash
+        // into the scalar field
+        let next_leaf = LegacyHash::hash(current_leaf.inner, current_leaf).into();
```
akirillo, line 188: @joeykraut I switched the implementation to use the STARK scalar field, the most important implication is this.

Cairo's Pedersen hash builtin operates over a felt hash state and outputs a felt (though it can take any type as input, i.e. `Scalar`). Since `Scalar` is a wrapper around a felt, we take the inner felt representation as the hash state for the Pedersen hash, and convert the felt output into a `Scalar` which, under the hood, reduces it to the scalar field.

This should be safe, but if you think of something I didn't, of course let me know. If it looks good, we'll of course have to be sure to match this implementation prover-side and in the integration tests

joeykraut: There are two problems as far as I see:
1. This would force us to use Pedersen hash prover side, I don't think we should
2. We would still have to implement nonnative arithmetic prover side to accommodate for the internal computation while computing the hash. Computing the hash in the scalar field and computing the hash in a larger field then reducing the result to the scalar field are not equivalent

PR #40, opened 2023-08-02
renegade_contracts: merkle: poseidon: Poseidon implementation over scalar field

akirillo:
This PR introduces a `PoseidonSponge` that computes the Poseidon hash over the STARK scalar field.

It mirrors and is fuzz tested against a reference implementation from Arkworks, albeit currently using dummy values for the number of rounds, round constants, and MDS matrix.

This PR includes some small refactors across the testing stack / utilities that cropped up, apologies for bundling them in here.

Integrating this hash with the Merkle contract is left for a future PR.


src/merkle/poseidon.cairo
```diff
@@ -0,0 +1,304 @@
...
+        let PoseidonSponge{mut state, mut mode, round_constants, mds } = self;
+
+        let round_constants_span = round_constants.deep_span();
+        let mds_span = mds.deep_span();
+
+        let mut absorb_index = match mode {
+            SpongeMode::Absorbing(i) => i,
+            SpongeMode::Squeezing(_) => {
+                permute(ref state, round_constants_span, mds_span);
+                0
```
joeykraut: do we need to be able to reset the sponge? We can just disallow this and simplify this code block if so

src/merkle/poseidon.cairo
```diff
@@ -0,0 +1,304 @@
...
+struct PoseidonSponge {
+    state: NullableVec<Scalar>,
+    mode: SpongeMode,
+    round_constants: Array<Array<Scalar>>,
+    mds: Array<Array<Scalar>>,
+}
+
+#[generate_trait]
+impl PoseidonImpl of PoseidonTrait {
+    fn new() -> PoseidonSponge {
```
joeykraut, line 50: Let's add doc comments on all these methods, for an outsider, `squeeze`, `absorb`, etc might not be immediately clear. Especially the way in which a sponge may switch between absorbing and squeezing

src/merkle/poseidon.cairo
```diff
@@ -0,0 +1,304 @@
...
+        mode = SpongeMode::Squeezing(squeeze_index);
+
+        self = PoseidonSponge { state, mode, round_constants, mds };
+
+        output
+    }
+}
+
+// DUMMY VALUES
+fn round_constants() -> Array<Array<Scalar>> {
```
joeykraut, line 162: What's the plan for this long term? Keep the arc/mds matrices in contract storage?

src/merkle/poseidon.cairo
```diff
@@ -0,0 +1,304 @@
...
+    ref state: NullableVec<Scalar>, round_constants: Span<Span<Scalar>>, mds: Span<Span<Scalar>>
+) {
+    let mut i = 0;
+    loop {
+        if i == FULL_ROUNDS {
+            break;
+        }
+
+        apply_round_constants(ref state, round_constants, i);
+        apply_sbox(ref state, true);
```
joeykraut: nit: add an inline comment describing the literal parameter `true` here, e.g. `true /* is_full_round */`

joeykraut: Same elsewhere as well

src/merkle/poseidon.cairo
```diff
@@ -0,0 +1,304 @@
...
+
+        let mut state_i = state[i];
+        state_i += *current_round_constants[i];
+        state.set(i, state_i);
+
+        i += 1;
+    };
+}
+
+fn apply_sbox(ref state: NullableVec<Scalar>, full: bool) {
```
joeykraut: nit: rename to `full_round` or similar, "full" is overloaded in the context of sponge constructions

joeykraut (approved):
lgtm, a few little things for readability

src/merkle/poseidon.cairo
akirillo: Honestly, I'd rather have them baked in as `const`s - once we compute the right round constants I can write a quick script to write them to a `.cairo` file

PR #42, opened 2023-08-03
renegade_contracts: darkpool: migrate to cairo v2.0.1

akirillo:
This PR migrates the darkpool contract to target v2.0.1 of the Cairo compiler.

This includes the necessary syntactic changes to the darkpool contract itself, as well as the subsumption of the modules previously defined in the `oz` folder, because there is no extensibility framework in this version of the compiler (discussion can be tracked [here](https://community.starknet.io/t/cairo-1-contract-syntax-is-evolving/94794)).

Porting integration tests is saved for a future PR.


src/darkpool.cairo
```diff
@@ -1,122 +1,582 @@
...
-        match_proof_blob: Array::<felt252>,
-        settle_proof_blob: Array::<felt252>,
-    ) -> felt252 {
-        DarkpoolLib::process_match(
-            party_0_payload, party_1_payload, match_proof_blob, settle_proof_blob, 
-        )
+    /// Executes a set of external ERC20 transfers
+    /// Parameters:
+    /// - `contract_address`: The address of the current contract // TODO: Should we just get this inside the body of this function?
+    /// - `caller_address`: The address triggering the withdrawal / deposit // TODO: Should we just get this inside the body of this function?
```
joeykraut: Re your `TODO`: No on `caller_address`, this is not usually the same address as the "caller" of `update_wallet` interface. The user will ERC20 approve a spend by the contract from their wallet then tell the relayer to go prove `VALID WALLET UPDATE` on their state transition. So the address of the deposit is the user's wallet, and the caller is the relayer's wallet. 

Maybe we should rename this field?

src/darkpool.cairo
```diff
@@ -1,122 +1,582 @@
...
+            } else {
+                // Withdraw
+                erc20.transfer(next_transfer.account_addr, next_transfer.amount);
+
+                // Emit event
+                self
+                    .emit(
+                        Event::Withdrawal(
+                            Withdrawal {
+                                recipient: caller_address,
```
joeykraut: The `sender` and `recipient` in these addresses should be from the `ExternalTransfer` struct

src/darkpool.cairo
```diff
@@ -1,122 +1,582 @@
...
+        assert(!initializable__is_initialized(@self), 'Initializable: is initialized');
+        self._initialized.write(true);
+    }
+
+    // ---------------
+    // | UPGRADEABLE |
+    // ---------------
+
+    #[external(v0)]
+    fn upgradeable__upgrade(ref self: ContractState, class_hash: ClassHash) {
```
joeykraut, line 581: Let's add doc comments to this method and any relevant others making explicit that the caller should handle access control

src/darkpool/types.cairo
```diff
@@ -0,0 +1,29 @@
...
+#[derive(Copy, Drop, Serde)]
+struct ExternalTransfer {
+    /// The address of the account contract to deposit from or withdraw to
+    account_addr: ContractAddress,
+    /// The mint (contract address) of the token being transferred
+    mint: ContractAddress,
+    /// The amount of the token transferred
+    amount: u256,
+    /// Whether or not the transfer is a deposit (otherwise a withdrawal)
+    is_deposit: bool,
```
joeykraut: I'd suggest we flip this to `is_withdrawal` to match the implementation of `ExternalTransferDirection` [here](https://github.com/renegade-fi/renegade/blob/main/circuits/src/types/transfers.rs#L35-L42). I.e. `ExternalTransferDirection::Deposit` serializes as `Scalar(0)` by default, so it'd be good to match this default behavior

src/darkpool/types.cairo
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/// Represents the artifacts produced by one of the parties in a match
+#[derive(Drop, Serde)]
+struct MatchPayload {
```
joeykraut, line 20: Love that we can have structured calldata now, really cleans up the interfaces, nice work 👍

src/oz/erc20.cairo
```diff
@@ -1 +1,20 @@
-mod interface;
+use starknet::ContractAddress;
+
+#[starknet::interface]
+trait IERC20<TContractState> {
```
joeykraut: Can we add a link to where this interface is standardized? I.e. is this based on an OpenZeppelin source impl?

joeykraut (approved):
A few small things, nothing blocking, really liking how this is all coming together, nice work!

PR #43, opened 2023-08-04
tests: darkpool: arkworks merkle reference tests

akirillo:
This PR ports over integration tests for the Darkpool against the Arkworks Merkle tree reference implementation onto the new testing stack.

Tests checking the functionality of `get_wallet_blinder_transaction`, `is_nullifier_used`, transfers, upgrades, and access controls in the native Cairo test runner is left for a future  PR.

A lot of this PR is setting up Rust types for `ExternalTransfer`, `MatchPayload`, and other basic starknet types, as well as a quick implementation of their serialization to calldata. This functionality will later be subsumed into the `starknet-client` crate once the crate refactor is stabilized.


tests/Cargo.toml
```diff
@@ -17,10 +17,12 @@ tokio = "1"
...
 ark-crypto-primitives = { version = "0.4", features = ["crh", "merkle_tree", "sponge"] }
 ark-std = "0.4"
+ark-ff = "0.4"
 rand = "0.8"
 num-bigint = { version = "0.4.3", features = ["rand"] }
 
 starknet = { workspace = true }
 katana-core = { git = "https://github.com/dojoengine/dojo.git" }
 dojo-test-utils = { git = "https://github.com/dojoengine/dojo.git" }
 mpc-stark = { git = "https://github.com/renegade-fi/mpc-stark.git" }
```
joeykraut, line 27: You may want to pin this to `v0.1.0`, although I'm going to re-release `mpc-stark` soon, so we can wait til then if you prefer. `main` is not necessarily stable

tests/src/darkpool/utils.rs
```diff
@@ -0,0 +1,288 @@
...
+    initialize(
+        account,
+        darkpool_address,
+        vec![merkle_class_hash, nullifier_set_class_hash, merkle_height],
+    )
+    .await
+    .map(|_| ())
+}
+
+pub async fn contract_get_wallet_blinder_transaction(
```
joeykraut: nit: naming is a bit odd, maybe just `get_wallet_blinder_transaction` and similar for the method below

tests/src/utils.rs
```diff
@@ -105,3 +107,247 @@ pub async fn invoke_contract(
         .map(|_| ())
         .map_err(|e| eyre!("Error invoking {}: {}", entry_point, e))
 }
+
+pub async fn contract_get_root(
```
joeykraut: same as before with the naming, maybe `get_contract_root` and `check_nullifier_used` below

tests/src/utils.rs
```diff
@@ -105,3 +107,247 @@ pub async fn invoke_contract(
...
+    )
+    .await
+    .map(|r| r[0] == FieldElement::ONE)
+}
+
+// ----------------
+// | MISC HELPERS |
+// ----------------
+
+pub fn scalar_to_felt(scalar: &Scalar) -> Result<FieldElement> {
```
joeykraut: This shouldn't practically error, we can just return `FieldElement` and call `expect` on the error

tests/src/utils.rs
```diff
@@ -105,3 +107,247 @@ pub async fn invoke_contract(
...
+        if self.is_identity() {
+            vec![FieldElement::ZERO, FieldElement::ZERO]
+        } else {
+            let aff = self.to_affine();
+            let x_bytes = aff.x.into_bigint().to_bytes_be();
+            let y_bytes = aff.y.into_bigint().to_bytes_be();
+            vec![
+                FieldElement::from_byte_slice_be(&x_bytes).unwrap(),
+                FieldElement::from_byte_slice_be(&y_bytes).unwrap(),
+            ]
```
joeykraut, line 274: can we just call `scalar_to_felt` here?

joeykraut (approved):
Few small things, nothing blocking

tests/Cargo.toml
akirillo: gotcha, let's wait then - I rely on the `Scalar::Field` associated type which doesn't seem to be present in 0.1.0

tests/src/darkpool/utils.rs
akirillo: yeahhh i started the naming and i rolled with it, also not sure about how i feel about it anymore lol, gonna drop the `contract_` prefix

tests/src/utils.rs
akirillo: hmm, not really - the affine coordinates are not of `Scalar` type, and I can't construct a `FieldElement` from them directly since its `inner` field is private

PR #44, opened 2023-08-04
renegade_contracts: testing: darkpool: cairo tests

akirillo:
This PR implements native Cairo runner sanity tests for the following darkpool functionality:
- Correct `last_modified` transaction hashes for wallets
- Correct nullifiers used
- Deposits / withdrawals correctly processed
- Upgrades correctly processed
- Access & initialization controls in place


src/darkpool.cairo
```diff
@@ -168,91 +182,103 @@ mod Darkpool {
...
-    /// - `nullifier_set_class_hash`: The hash of the implementation class used for nullifier set operations
-    #[external(v0)]
-    fn upgrade_nullifier_set(ref self: ContractState, nullifier_set_class_hash: ClassHash) {
-        ownable__assert_only_owner(@self);
+        /// Upgrades the Darkpool implementation class
+        /// Parameters:
+        /// - `darkpool_class_hash`: The hash of the new implementation class
+        fn upgrade(ref self: ContractState, darkpool_class_hash: ClassHash) {
+            ownable__assert_only_owner(@self);
+            upgradeable__upgrade(ref self, darkpool_class_hash);
```
joeykraut, line 241: We should emit an event here like we do for other upgrades

src/testing/tests/darkpool_tests.cairo
```diff
@@ -0,0 +1,598 @@
...
+
+#[test]
+#[available_gas(1000000000)] // 10x
+fn test_upgrade_darkpool() {
+    let test_caller = contract_address_try_from_felt252(TEST_CALLER).unwrap();
+    set_contract_address(test_caller);
+    let mut darkpool = setup_darkpool();
+
+    darkpool.upgrade(DummyUpgradeTarget::TEST_CLASS_HASH.try_into().unwrap());
+    assert(
```
joeykraut: can we add a comment here referencing that the `DummyUpgradeTarget` has a hardcoded response for this call? It's a bit contextualized when reading this test

src/testing/tests/darkpool_tests.cairo
```diff
@@ -0,0 +1,598 @@
...
+    assert(darkpool_balance == INIT_BALANCE - TRANSFER_AMOUNT, 'incorrect darkpool balance');
+}
+
+// -----------------
+// | UPGRADE TESTS |
+// -----------------
+
+#[test]
+#[available_gas(1000000000)] // 10x
+fn test_upgrade_darkpool() {
```
joeykraut, line 327: Can we add an analogous test in which we maybe go: darkpool (v1) -> dummy (v2) -> darkpool (v3), and verify that storage is preserved? I.e. spent nullifiers or `wallet_last_updated` is preserved from version 1 to version 3.

joeykraut (approved):
Looks good, nice abstractions

src/darkpool.cairo
akirillo: we emit an event in the call to `upgradeable__upgrade`, we do so there because the implementation is taken from OpenZeppelin, and I want to minimize the diff for when they (finally) release their contracts via Scarb

akirillo:
Will merge after #43

PR #47, opened 2023-08-07
tests, renegade_contracts: split out transcript tests

akirillo:
This PR splits out transcript tests against the devnet node, doing a single basic fuzz test. Also includes all the relevant test scaffolding & utilities for this.

### DIFF
--- a/tests/src/transcript/utils.rs
+++ b/tests/src/transcript/utils.rs
@@ -1,0 +1,197 @@
+// ---------------------
+// | META TEST HELPERS |
+// ---------------------
+
+use dojo_test_utils::sequencer::TestSequencer;
+use eyre::Result;
+use merlin::HashChainTranscript;
+use mpc_stark::algebra::{scalar::Scalar, stark_curve::StarkPoint};
+use once_cell::sync::OnceCell;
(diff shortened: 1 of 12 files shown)


tests/src/transcript/utils.rs
```diff
@@ -0,0 +1,205 @@
...
+    ];
+    deploy(account, class_hash, &calldata).await?;
+    Ok(calculate_contract_address(class_hash, &calldata))
+}
+
+// --------------------------------
+// | CONTRACT INTERACTION HELPERS |
+// --------------------------------
+
+pub async fn contract_rangeproof_domain_sep(account: &ScriptAccount, n: u64, m: u64) -> Result<()> {
```
joeykraut: nit: Naming is a bit weird here, maybe we can name it something like what we did in the other tests? E.g. `call_rangeproof_domain_sep` and similar for the methods below?

joeykraut (approved):

tests/src/transcript/utils.rs
akirillo: whoops, broke this branch off before that refactor, will do so

PR #49, opened 2023-08-08
tests: verifier_utils: integration test scaffolding

akirillo:
This PR introduces scaffolding and utilities to be used in integration tests of the verifier utilities, namely the `calc_delta`, `get_s_elem`, and `squeeze_challenge_scalars` Cairo functions which will be tested against their counterparts in `mpc-bulletproof`.

This includes a wrapper contract that exposes these utility functions, and helpers for getting their exact inputs from the `mpc-bulletproof` implementation (e.g., no single method that "gets challenge scalars for given proof").

This also includes an implementation of the dummy circuit used in the native Cairo tests.

### DIFF
--- a/src/testing/tests/utils_tests.cairo
+++ b/src/testing/tests/utils_tests.cairo
@@ -62,6 +62,8 @@ fn test_flatten_sparse_weight_matrix_basic() {
     let z = 2.into();
     let width = 4;
 
+    // For z := [2, 2^2, 2^4, ...] we have expected := zW
+    // ("flattening" W matrix via left-multiplication by z)
     let mut expected = ArrayTrait::new();
     // 2*1 + 4*2 + 8*4 = 42
     expected.append(42.into());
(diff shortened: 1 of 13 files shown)


src/testing/tests/utils_tests.cairo
```diff
@@ -62,8 +62,6 @@ fn test_flatten_sparse_weight_matrix_basic() {
     let z = 2.into();
     let width = 4;
 
-    // For z := [2, 2^2, 2^4, ...] we have expected := zW
-    // ("flattening" W matrix via left-multiplication by z)
```
joeykraut: any reason we deleted this comment, or was this by mistake?

joeykraut (approved):
lgtm

src/testing/tests/utils_tests.cairo
akirillo: ah, got caught up in a rebase it seems, my b

PR #50, opened 2023-08-08
tests: verifier_utils: implement verifier utils tests

akirillo:
This PR builds on top of #49 to implement tests for `squeeze_challenge_scalars`, `get_s_elem`, and `calc_delta` against reference implementations from `mpc-bulletproof`.

Depends on https://github.com/renegade-fi/mpc-bulletproof/pull/9


tests/src/verifier_utils/utils.rs
```diff
@@ -149,11 +154,11 @@ pub async fn squeeze_challenge_scalars(
...
         let mut r_iter = r.iter();
 
-        let challenge_scalars_len = r_iter
-            .next()
-            .unwrap()
-            .to_bytes_be()
-            .as_slice()
+        let mut challenge_scalars_len_cursor = Cursor::new(r_iter.next().unwrap().to_bytes_be());
+        // Grab the least signifcant 4 bytes for the len u32
+        challenge_scalars_len_cursor.set_position(28);
```
joeykraut, line 159: I feel the cursor approach obscures what is going on here, especially since we seek to `28` bytes in. What was wrong with the previous approach of casting the bytes to a slice and reading that way? I prefer that

tests/src/verifier_utils/utils.rs
```diff
@@ -163,16 +168,140 @@ pub async fn squeeze_challenge_scalars(
...
-        let u_len = r_iter
-            .next()
-            .unwrap()
-            .to_bytes_be()
-            .as_slice()
-            .read_u32::<BigEndian>()
-            .unwrap() as usize;
+        let mut u_len_cursor = Cursor::new(r_iter.next().unwrap().to_bytes_be());
+        // Grab the least signifcant 4 bytes for the len u32
+        u_len_cursor.set_position(28);
```
joeykraut, line 173: same as above

joeykraut (approved):

tests/src/verifier_utils/utils.rs
akirillo: It was actually reading incorrectly - since it's big-endian, `read_u32::<BigEndian>` would read the first 4 bytes instead of the last 4, always returning 0. The most I can do with the `ReadBytesExt` trait from `byteorder` is a `read_u128`, so I'd have to call that twice to get to the actual non-zero bytes. At that point, explicitly setting the position to 28 feels clearer.

PR #51, opened 2023-08-08
tests: verifier: scaffolding for verifier integration tests

akirillo:
This PR introduces scaffolding & utilities for running verifier integration tests (attempting verification of valid & invalid proofs), as well as slightly tweaking the verifier contract interface to only return the `verified` field of the `VerificationJob`, as this is all that is relevant.

### DIFF
--- a/tests/src/utils.rs
+++ b/tests/src/utils.rs
@@ -260,17 +260,29 @@ impl MatchPayload {
 }
 
 pub struct CircuitParams {
+    /// Number of multiplication gates in the circuit
     pub n: usize,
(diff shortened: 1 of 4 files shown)


tests/src/utils.rs
```diff
@@ -259,13 +259,36 @@ impl MatchPayload {
     }
 }
 
+pub struct CircuitParams {
```
joeykraut, line 262: nit: Can we add doc comments for each field? I know it's documented elsewhere but would be nice to have

joeykraut (approved):
Lgtm

PR #53, opened 2023-08-09
renegade_contracts: utils: storage: arbitrarily large object storage

akirillo:
This PR switches out the implementation of the `StorageAccessSerdeWrapper` to be able to write arbitrarily large objects to storage. This is still done by serializing the object first, but the serialization is now split across many 256-slot storage segments following [alexandria's `List` type implementation](https://github.com/keep-starknet-strange/alexandria/tree/cairo-v2.0.1/src/storage).

### DIFF
--- a/src/utils/storage.cairo
+++ b/src/utils/storage.cairo
@@ -8,12 +8,13 @@ use ec::EcPoint;
 use starknet::{
     StorageAccess, SyscallResult, SyscallResultTrait,
     storage_access::{
-        StorageBaseAddress, storage_address_from_base, storage_address_from_base_and_offset,
+        StorageAddress, StorageBaseAddress, storage_address_from_base,
+        storage_address_from_base_and_offset,
     },
     syscalls::{storage_read_syscall, storage_write_syscall}
 };
 
-use alexandria::data_structures::array_ext::ArrayTraitExt;
+use alexandria::{data_structures::array_ext::ArrayTraitExt, storage::list::{List, ListTrait}};
 
 use super::serde::EcPointSerde;
 
@@ -23,72 +24,33 @@ use super::serde::EcPointSerde;
 // StorageAccess implementations for some types.
 #[derive(Drop)]
 struct StorageAccessSerdeWrapper<T> {
-    inner: T
+    inner: List<felt252>, 
 }
 
-impl StorageAccessSerdeImpl<T, impl TSerde: Serde<T>> of Serde<StorageAccessSerdeWrapper<T>> {
-    fn serialize(self: @StorageAccessSerdeWrapper<T>, ref output: Array<felt252>) {
-        let mut inner_output = ArrayTrait::new();
-        self.inner.serialize(ref inner_output);
-        output.append(inner_output.len().into());
-        output.append_all(ref inner_output);
-    }
-
-    fn deserialize(ref serialized: Span<felt252>) -> Option<StorageAccessSerdeWrapper<T>> {
-        serialized.pop_front()?;
-        let inner = Serde::<T>::deserialize(ref serialized)?;
-        Option::Some(StorageAccessSerdeWrapper { inner })
-    }
-}
-
-impl StorageSerdeImpl<
-    T, impl Tserde: Serde<T>, impl TDrop: Drop<T>
-> of StorageAccess<StorageAccessSerdeWrapper<T>> {
+// Would love to #[derive(storage_access::StorageAccess)] on `StorageAccessSerdeWrapper<T>`,
+// but the compiler doesn't properly derive the `StorageAccess` trait for generic types.
+// So, we manually call out to the `StorageAccess` impl of the underlying `List` type.
+impl WrapperStorageAccessImpl<T> of StorageAccess<StorageAccessSerdeWrapper<T>> {
     fn read(
         address_domain: u32, base: StorageBaseAddress
     ) -> SyscallResult<StorageAccessSerdeWrapper<T>> {
-        StorageSerdeImpl::<T>::read_at_offset_internal(address_domain, base, 0)
-    }
-
-    fn read_at_offset_internal(
-        address_domain: u32, base: StorageBaseAddress, offset: u8
-    ) -> SyscallResult<StorageAccessSerdeWrapper<T>> {
-        // Read serialization len, add 1 to account for the len slot
-        let num_slots: u8 = storage_read_syscall(address_domain, storage_address_from_base(base))?
-            .try_into()
-            .expect('Storage - value too large')
-            + 1_u8;
-
-        let mut serialized = ArrayTrait::new();
-        let mut slots_read: u8 = 0;
-        loop {
-            if slots_read == num_slots {
-                break;
-            }
-
-            let address = storage_address_from_base_and_offset(base, offset + slots_read);
-            // Unwrapping here for now since you can't return / use `?` in a loop
-            let felt = storage_read_syscall(address_domain, address).unwrap_syscall();
-            serialized.append(felt);
-            slots_read += 1;
-        };
-        let mut serialized_span = serialized.span();
-
-        // This deserialize expects to pop off the len slot
-        match Serde::<StorageAccessSerdeWrapper<T>>::deserialize(ref serialized_span) {
-            Option::Some(val) => Result::Ok(val),
-            Option::None(_) => {
-                let mut data = ArrayTrait::new();
-                data.append('Storage - deser failed');
-                Result::Err(data)
-            },
-        }
+        let inner = StorageAccess::<List<felt252>>::read(address_domain, base)?;
+        Result::Ok(StorageAccessSerdeWrapper { inner })
     }
 
     fn write(
         address_domain: u32, base: StorageBaseAddress, value: StorageAccessSerdeWrapper<T>
     ) -> SyscallResult<()> {
-        StorageSerdeImpl::<T>::write_at_offset_internal(address_domain, base, 0, value)
+        StorageAccess::<List<felt252>>::write(address_domain, base, value.inner)
+    }
+
+    fn read_at_offset_internal(
+        address_domain: u32, base: StorageBaseAddress, offset: u8
+    ) -> SyscallResult<StorageAccessSerdeWrapper<T>> {
+        let inner = StorageAccess::<List<felt252>>::read_at_offset_internal(
+            address_domain, base, offset
+        )?;
+        Result::Ok(StorageAccessSerdeWrapper { inner })
     }
 
     fn write_at_offset_internal(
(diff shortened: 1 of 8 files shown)


src/utils/storage.cairo
```diff
@@ -23,72 +24,33 @@ use super::serde::EcPointSerde;
 // StorageAccess implementations for some types.
 #[derive(Drop)]
 struct StorageAccessSerdeWrapper<T> {
```
joeykraut: Do we still need the generic here?

src/utils/storage.cairo
```diff
@@ -23,72 +24,33 @@ use super::serde::EcPointSerde;
...
-    }
-}
-
-impl StorageSerdeImpl<
-    T, impl Tserde: Serde<T>, impl TDrop: Drop<T>
-> of StorageAccess<StorageAccessSerdeWrapper<T>> {
+// Would love to #[derive(storage_access::StorageAccess)] on `StorageAccessSerdeWrapper<T>`,
+// but the compiler doesn't properly derive the `StorageAccess` trait for generic types.
+// So, we manually call out to the `StorageAccess` impl of the underlying `List` type.
+impl WrapperStorageAccessImpl<T> of StorageAccess<StorageAccessSerdeWrapper<T>> {
```
joeykraut: This is too complicated and fragile, we should iterate on this before merging, or at least convince ourselves there is _truly_ no simpler way to accomplish this.

Why can't we just implement `StorageAccess<StorageAccessSerdeWrapper<T>>` as:
1. Serialize to `Array<felt252>`
2. Convert this array to `List<felt252>`
3. Delegate the `StorageAccess` methods to those implemented on `List`

Essentially I don't see the need for all this `unwrap` and `rewrap` logic -- we should just hang onto the runtime object at runtime and only use the `List` as an intermediate interface with storage

src/utils/storage.cairo
```diff
@@ -97,35 +59,70 @@ impl StorageSerdeImpl<
...
+    /// This "unwraps" the value being stored by reading its serialization from storage & deserializing it.
+    fn unwrap(self: StorageAccessSerdeWrapper<T>) -> T {
+        let mut serialized_span = self.inner.array().span();
+        let inner: T = Serde::deserialize(ref serialized_span).unwrap();
+        inner
+    }
+
+    /// This "rewraps" the value being stored by overwriting its serialization in storage.
+    /// IMPORTANT: THIS ASSUMES THAT `.write` IS CALLED WITH THE `StorageAccessSerdeWrapper` AFTER THIS FUNCTION IS CALLED.
+    /// See the doc comments below on why this is necessary.
```
joeykraut: Which doc comment? The order of these comments is confusing, let's just repaste here at risk of redundancy

src/utils/storage.cairo
```diff
@@ -97,35 +59,70 @@ impl StorageSerdeImpl<
...
+> of StorageAccessSerdeTrait<T> {
+    /// This "unwraps" the value being stored by reading its serialization from storage & deserializing it.
+    fn unwrap(self: StorageAccessSerdeWrapper<T>) -> T {
+        let mut serialized_span = self.inner.array().span();
+        let inner: T = Serde::deserialize(ref serialized_span).unwrap();
+        inner
+    }
+
+    /// This "rewraps" the value being stored by overwriting its serialization in storage.
+    /// IMPORTANT: THIS ASSUMES THAT `.write` IS CALLED WITH THE `StorageAccessSerdeWrapper` AFTER THIS FUNCTION IS CALLED.
```
joeykraut: This is a very tight coupling with the caller, if there is no way to refactor the access pattern such that we avoid `unwrap`/`rewrap` calls, then we should add a layer of indirection here to group these calls together and remove this assumption

src/utils/storage.cairo
```diff
@@ -97,35 +59,70 @@ impl StorageSerdeImpl<
...
+    /// See the doc comments below on why this is necessary.
+    fn rewrap(mut self: StorageAccessSerdeWrapper<T>, value: T) -> StorageAccessSerdeWrapper<T> {
```
joeykraut: It seems that all this is to the end of modifying the `StorageAcessSerdeWrapper` implementor in place, why is this desired? We seem to exclusively use it by just writing over the old value, which to me suggests no need for the rewrap, in-place flow

joeykraut (commented):
Let's think about other approaches to tame complexity here, I'll do a full review once we have settled on an approach

src/utils/storage.cairo
akirillo: Technically no, but I think it makes the contract code significantly more readable to see e.g. `StorageAccessSerdeWrapper<VerificationJob>` in the `Storage` struct as opposed to just a `StorageAccessSerdeWrapper`

src/utils/storage.cairo
akirillo: I don't see this block of code as being fragile, really - it mimics what the implementation of `#[derive(storage_access::StorageAccess)]` would do here by delegating the methods to those implemented on `List`, as you say. The only reason I can't do that is because `#[derive(storage_access::StorageAccess)]` doesn't expand generics properly - to your earlier comment, we could drop the generic parameter from the `StorageAccessSerdeWrapper<T>` type to be able to `derive` but I am strongly in favor of leaving it for readability.

It's worth calling out here that it is not the `StorageAccess` methods on `List` that implement writing all of the list elements to storage. The `StorageAccess` trait is quite opinionated and is intended only for writing objects smaller than 256 slots to storage. `List` works around this by calculating `StorageAddress`es using a hash of the base address & a counter, and it's actual `StorageAccess` trait implementation is only concerned with storing a single `u32`, the list's length, to the `base` `StorageAddress` as passed into the `StorageAccess` trait methods.

So it's not quite straightforward to implement this logic inside the `StorageAccess` trait methods themselves, but more so I think it's somewhat of an abstraction boundary violation, and the stark difference in access pattern here (reading / writing across multiple 256-slot chunks) should be encapsulated within a separate trait, hence the `unwrap` and `rewrap` logic.

I'm definitely open to iterating on this, let's do it live - but I am quite happy with the tradeoff in complexity, ergonomics, and cost here - there is a small amount of redundant reads/writes that we can optimize out if we fork the `List` implementation.

src/utils/storage.cairo
akirillo: Unfortunately not - the `write` call indicated here is actually called on a type called `ContractState`, which is expanded by the compiler only within the scope of a contract (it is a different type per-contract). I would rather not re-write this trait in the scope of any contract that may use it, and I haven't been able to find the trait exposing the `read` and `write` methods on the `ContractState` types that I could import - if I could, then I could accept a generic parameter here with that trait bound and call `write` within this method itself, as you suggest.

src/utils/storage.cairo
joeykraut: If there is no way around this then we should remove this abstraction and implement storage access directly on the wrapped types, there's no point in an abstraction that increases complexity, especially not for DRY. And we don't really use this abstraction many places

src/utils/storage.cairo
akirillo: The issue is that it's impossible to construct the underlying `List` struct in the runtime, since its members include the `address_domain` and `base` (`StorageAddressBase`) that the list was accessed at in storage. These are the same arguments you see in the `StorageAccess` trait methods above. However, these trait methods are called under-the-hood in the `read` and `write` methods on the `ContractState` type, so there's no place where we can access or derive these values to construct the `List` manually.

Thus the pattern I came up with is to first read the `StorageAccessSerdeWrapper` from storage using the high-level `read` method on `ContractState`, which automatically populates these fields on the underlying `List` (this is done in the implementation of `StorageAccess<List>` [here](https://github.com/keep-starknet-strange/alexandria/blob/cairo-v2.0.1/src/storage/list.cairo#L195-L199)). This lends itself to e.g. the following access pattern: `contract.my_storage_access_wrapper.read().unwrap()`

src/utils/storage.cairo
akirillo: I tend to disagree here - many of the types are at risk of becoming too big for the intended sub-256 slot storage access pattern because they hold arrays of complex types. The best way to implement storage access directly on them would still be the same high level approach, serialize to an array of felts and then spread that across multiple 256-slot chunks. I don't see the implementation differing substantially, and this way we have a straightforward way to store any type as long as it implements `Serde`, which is one of the few derivable traits

tests/src/utils.rs
```diff
@@ -141,8 +141,8 @@ pub async fn is_nullifier_used(
 // ----------------
 
 pub fn random_felt() -> FieldElement {
-    let modulus = BigUint::from_bytes_be(&FieldElement::MAX.to_bytes_be()) + 1_u8;
-    let rand_biguint = thread_rng().gen_biguint_below(&modulus);
+    let max = BigUint::from_bytes_be(&FieldElement::MAX.to_bytes_be());
```
akirillo, line 144: saw this trip up a test in CI so decided to toss the change in here since it's so small

src/utils/storage.cairo
```diff
@@ -97,35 +53,115 @@ impl StorageSerdeImpl<
...
+        1
     }
 }
+
+// Implementation (and documentation) taken from: https://github.com/keep-starknet-strange/alexandria/blob/cairo-v2.0.1/src/storage/list.cairo
+// This function finds the StorageBaseAddress of a "storage segment" (a continuous space of 256 storage slots)
+// and an offset into that segment where a value at `index` is stored
+// each segment can hold up to 256 felts
+//
+// the way how the address is calculated is very similar to how a LegacyHash map works:
```
joeykraut: I know it's a copy-paste
```suggestion
// the way the address is calculated is very similar to how a LegacyHash map works:
```

src/utils/storage.cairo
```diff
@@ -97,35 +53,115 @@ impl StorageSerdeImpl<
...
+//
+// first we take the `list_base` address which is derived from the name of the storage variable
+// then we hash it with a `key` which is the number of the segment where the element at `index` belongs (from 0 upwards)
+// we hash these two values: H(list_base, key) to the the `segment_base` address
+// finally, we calculate the offset into this segment, taking into account the size of the elements held in the array
```
joeykraut: I know this is copy-paste but can you improve this documentation, I feel you'll do a better job with the wording

src/utils/storage.cairo
```diff
@@ -97,35 +53,115 @@ impl StorageSerdeImpl<
...
+// etc.
+//
+// where addresses of each segment are:
+//
+// segment0 = hash(bar.address(), 0)
+// segment1 = hash(bar.address(), 1)
+// segment2 = hash(bar.address(), 2)
+//
+// so for getting a Foo at index 90 this function would return address of segment2 and offset of 26
+fn calculate_base_and_offset_for_index(
```
joeykraut, line 111: Can you explain (and perhaps document it it's not obvious) why we don't just call out to the underlying implementation, i.e. why do we need to copy it here? Can we not just construct these objects with the correct pointers at runtime and then call out to their implementation? We definitely know the base address at runtime, and the segment computation can all be done at runtime

This would be preferable because it places us downstream of any optimizations the alexandria team makes

src/utils/storage.cairo
```diff
@@ -97,35 +53,115 @@ impl StorageSerdeImpl<
...
+    list_base: StorageBaseAddress, index: u32
+) -> (StorageBaseAddress, u8) {
+    let (key, offset) = U32DivRem::div_rem(index, MAX_STORAGE_SEGMENT_ELEMS.try_into().unwrap());
```
joeykraut: nit: prefer `(segment_index, offset)` for readability. It's okay to deviate from the naming in alexandria if it improves readability

tests/src/utils.rs
akirillo: this is somehow still panicking, but rarely flakes out so will handle later

joeykraut (approved):
This looks a lot better, really well done. Definitely a hard, complex bit of code to provide a clean interface to, and I like where we've ended up

PR #54, opened 2023-08-10
renegade_contracts: merkle: poseidon: proper STARK scalar field Poseidon params

akirillo:
This PR ports over the properly-generated Poseidon hash parameters over the STARK scalar field, which can be found in the [relayer repo](https://github.com/renegade-fi/renegade/blob/joey/crate-refactor/renegade-crypto/src/constants.rs).

It also updates the Poseidon tests to directly pull these parameters from the `renegade-crypto` crate in the relayer repo


src/merkle/poseidon/constants.cairo
```diff
@@ -0,0 +1,264 @@
...
+    mds.append(mds_0);
+    mds.append(mds_1);
+    mds.append(mds_2);
+
+    mds
+}
+
+
+// DUMMY VALUES
+// TODO: Hardcode all round constants
```
joeykraut: remove these, instead let's link to the relayer repo or copy the comment on how they were generated

joeykraut (approved):
lgtm

PR #56, opened 2023-08-10
renegade_contracts: darkpool: initialize darkpool w/ single verifier

akirillo:
This PR is the beginning of the gradual integration of the verifier into the darkpool.

Here, we add the verifier to the darkpool initialization process, for now initializing just a single verifier contract (in the future, there will be multiple verifier contracts for each of the circuits).

We do not yet actually invoke the verifier contract (other than initializing it). Implementing access controls for the verifier contract is also saved for a future PR.

This PR also makes the relevant changes to the integration testing code to ensure that the darkpool is initialized with a verifier.

Since there are a lot of shared utilities between the integration testing code and the deploy/upgrade scripts, the minimal reasonable changes are made to the latter, as well.

However, updating the deploy / upgrade scripts to properly include the verifier contract is left for a future PR.

**Testing:**
Cairo tests pass, and integration tests that are affected by these changes (`darkpool`, `verifier`, `verifier_utils`) pass.

### DIFF
--- a/src/testing/test_utils.cairo
+++ b/src/testing/test_utils.cairo
@@ -2,9 +2,11 @@ use array::ArrayTrait;
 use serde::Serde;
 use option::OptionTrait;
 use traits::Into;
-use ec::{ec_point_from_x, ec_mul};
+use ec::{ec_point_from_x, ec_mul, ec_point_new, StarkCurve};
 
-use renegade_contracts::verifier::types::{Proof, SparseWeightMatrix};
+use renegade_contracts::verifier::types::{
+    Proof, SparseWeightMatrix, SparseWeightVec, CircuitParams
+};
 
 const DUMMY_ROOT_INNER: felt252 = 'DUMMY_ROOT';
 const DUMMY_WALLET_BLINDER_TX: felt252 = 'DUMMY_WALLET_BLINDER_TX';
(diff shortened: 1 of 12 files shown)


src/testing/test_utils.cairo
akirillo, line 1: `get_dummy_circuit_params` and other helpers here were copied from git history, when we used to do verifier tests in the cairo test runner before the integration tests were implemented.

tests/src/utils.rs
```diff
@@ -437,3 +416,171 @@ impl CalldataSerializable for CircuitParams {
             .collect()
     }
 }
+
```
akirillo, line 419: All the helpers below were moved from `tests/src/verifier.utils`

starknet_scripts/src/commands/deploy.rs
```diff
@@ -7,8 +7,8 @@ use tracing::{debug, info};
 use crate::{
     cli::{Contract, DeployArgs},
     commands::utils::{
-        deploy_darkpool, deploy_merkle, deploy_nullifier_set, initialize, setup_account,
-        MERKLE_HEIGHT,
+        deploy_darkpool, deploy_merkle, deploy_nullifier_set, deploy_verifier, initialize,
```
joeykraut, line 10: For future reference, I think separating out the `starknet_scripts` changes into a separate PR would be good

joeykraut (approved):

PR #57, opened 2023-08-12
renegade_contracts: darkpool: add verification polling methods

akirillo:
This PR introduces "polling" methods accompanying `new_wallet`, `update_wallet`, and `process_match`, which call out to the underlying `verifier.step_verification` method and, when it completes, handle the merkle / nullifier set updating logic (should the proof verify).

As such, calling `new_wallet` / etc. will simply enqueue the verification jobs, and to drive them forward, one must call `poll_new_wallet` / etc.

A deliberate unknown left here is whether or not enqueueing the 6 verification jobs necessary for `process_match` will be possible within a single transaction. For full size circuits, almost definitely not, but this is something I intend to iterate on, for now optimistically assuming that it will work.

The darkpool Cairo tests are now excluded from the `testing` module since they will all need to be ported over to e2e tests (since they now invoke the verifier, they will need a proof generated from `mpc-bulletproof` as input).

The e2e test updates are coming in a separate PR.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -137,13 +137,16 @@ mod Darkpool {
         nullifier_set_class_hash: ClassHash,
         /// Stores the contract address of the deployed verifier
         verifier_contract_address: ContractAddress,
-        /// Mapping of callback elements for in-progress `new_wallet` verification jobs
+        /// Mapping of elements to be used in the post-polling merkle & nullifier set
+        /// callback logic for in-progress `new_wallet` verification jobs
         new_wallet_callback_elems: LegacyMap<felt252,
         StorageAccessSerdeWrapper<NewWalletCallbackElems>>,
-        /// Mapping of callback elements for in-progress `update_wallet` verification jobs
+        /// Mapping of elements to be used in the post-polling merkle & nullifier set
+        /// callback logic for in-progress `update_wallet` verification jobs
         update_wallet_callback_elems: LegacyMap<felt252,
         StorageAccessSerdeWrapper<UpdateWalletCallbackElems>>,
-        /// Mapping of callback elements for in-progress `process_match` verification jobs
+        /// Mapping of elements to be used in the post-polling merkle & nullifier set
+        /// callback logic for in-progress `process_match` verification jobs.
         /// Uses the first verification job id in the list of ids for the process_match proofs
         /// as the mapping key.
         process_match_callback_elems: LegacyMap<felt252,
@@ -456,8 +459,6 @@ mod Darkpool {
         /// Poll the new wallet verification job, and if it verifies, insert the new wallet into the
         /// merkle tree
         /// Parameters:
-        /// - `wallet_blinder_share`: The public share of the wallet blinder, used for indexing
-        /// - `wallet_share_commitment`: The commitment to the new wallet's shares
         /// - `verification_job_id`: The ID of the verification job to step through
         /// Returns:
         /// - The new root after the wallet is inserted into the tree, if the proof verifies
@@ -512,14 +513,9 @@ mod Darkpool {
             _poll_update_wallet_inner(ref self, callback_elems, verification_job_id)
         }
 
-        /// Poll the update wallet verification job, and if it verifies, insert the new wallet into the
+        /// Poll the update wallet verification job, and if it verifies, insert the updated wallet into the
         /// merkle tree
         /// Parameters:
-        /// - `wallet_blinder_share`: The public share of the wallet blinder, used for indexing
-        /// - `wallet_share_commitment`: The commitment to the updated wallet's shares
-        /// - `old_shares_nullifier`: The nullifier for the public shares of the wallet before it was updated
-        /// - `public_wallet_shares`: The public shares of the wallet after it was updated
-        /// - `external_transfers`: The external transfers (ERC20 deposit/withdrawal)
         /// - `verification_job_id`: The ID of the verification job to step through
         /// Returns:
         /// - The root of the tree after the new commitment is inserted, if the proof verifies
@@ -612,6 +608,12 @@ mod Darkpool {
             _poll_process_match_inner(ref self, callback_elems, verification_job_ids, )
         }
 
+        /// Poll the process match verification job, and if it verifies, insert the updated wallet into the
+        /// merkle tree
+        /// Parameters:
+        /// - `verification_job_id`: The ID of the verification job to step through
+        /// Returns:
+        /// - The root of the tree after the new commitment is inserted, if the proof verifies
         fn poll_process_match(
             ref self: ContractState, verification_job_ids: Array<felt252>, 
         ) -> Option<Result<Scalar, felt252>> {
@@ -753,6 +755,8 @@ mod Darkpool {
         match verified {
             Option::Some(success) => {
                 if success {
+                    // Callback logic
+
                     // Insert the new wallet's commitment into the Merkle tree
                     let merkle_tree = _get_merkle_tree(@self);
                     let new_root = merkle_tree.insert(callback_elems.wallet_share_commitment);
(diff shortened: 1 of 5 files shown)


src/darkpool.cairo
```diff
@@ -49,7 +53,14 @@ trait IDarkpool<TContractState> {
         public_wallet_shares: Array<Scalar>,
         proof: Proof,
         witness_commitments: Array<EcPoint>,
-    ) -> Scalar;
+        verification_job_id: felt252,
+    ) -> Option<Result<Scalar, felt252>>;
+    fn poll_new_wallet(
+        ref self: TContractState,
+        wallet_blinder_share: Scalar,
+        wallet_share_commitment: Scalar,
```
joeykraut: Why do we need these calldata elements? They seem redundant, and we should aim to keep our interface lean

joeykraut: same below

src/darkpool.cairo
```diff
@@ -594,6 +747,144 @@ mod Darkpool {
...
+
+        match verified {
+            Option::Some(success) => {
+                if success {
+                    // Insert the new wallet's commitment into the Merkle tree
+                    let merkle_tree = _get_merkle_tree(@self);
+                    let new_root = merkle_tree.insert(wallet_share_commitment);
+
+                    // Mark wallet as updated
+                    _mark_wallet_updated(ref self, wallet_blinder_share);
```
joeykraut: I see why you added these to calldata. We should use the ones in the proof. One security check that's missing here is that these calldata commitments and blinder shares match those given in the proof. Instead, let's avoid needing this check and use the fields on the proof directly.

src/darkpool.cairo
```diff
@@ -594,6 +747,144 @@ mod Darkpool {
...
+        match verified {
+            Option::Some(success) => {
+                if success {
+                    // Insert the updated wallet's commitment into the Merkle tree
+                    let merkle_tree = _get_merkle_tree(@self);
+                    let new_root = merkle_tree.insert(wallet_share_commitment);
+
+                    // Add the old shares nullifier to the spent nullifier set
+                    let nullifier_set = _get_nullifier_set(@self);
+                    nullifier_set.mark_nullifier_used(old_shares_nullifier);
```
joeykraut: Same as above, we should *not* be using the raw calldata.

src/darkpool.cairo
```diff
@@ -594,6 +747,144 @@ mod Darkpool {
...
+        match valid_settle_verified {
+            Option::Some(success) => {
+                if success {
+                    let nullifier_set = _get_nullifier_set(@self);
+                    nullifier_set.mark_nullifier_used(party_0_old_shares_nullifier);
+                    nullifier_set.mark_nullifier_used(party_1_old_shares_nullifier);
+
+                    let merkle_tree = _get_merkle_tree(@self);
+                    merkle_tree.insert(party_0_wallet_share_commitment);
+                    let new_root = merkle_tree.insert(party_1_wallet_share_commitment);
```
joeykraut: calling out here the same as above for completeness

joeykraut (dismissed):
Structurally looks good, a few bugs to clear up

src/darkpool.cairo
akirillo: Hmm... so in that case, we'd be inserting the witness commitments submitted alongside the proof into the merkle tree and nullifier set (as opposed to plaintext public inputs), right? I can definitely grab those out of `witness_commitments` and remove these calldata elements from all of the interfaces, but I think we'd still need a check that those witness commitments are the same ones submitted to the verifier. I'm not sure the best way to do this, may take a bit more design (and prob just using a bit more contract storage), which I can do and then put this PR back up.

Does this track / am I misunderstanding?

akirillo:
@joeykraut will fix formatting + squash after review, if i do this now then the rebase will bork your ability to review just the last commit

src/darkpool.cairo
```diff
@@ -113,6 +137,17 @@ mod Darkpool {
...
+        new_wallet_callback_elems: LegacyMap<felt252,
+        StorageAccessSerdeWrapper<NewWalletCallbackElems>>,
+        /// Mapping of callback elements for in-progress `update_wallet` verification jobs
+        update_wallet_callback_elems: LegacyMap<felt252,
+        StorageAccessSerdeWrapper<UpdateWalletCallbackElems>>,
+        /// Mapping of callback elements for in-progress `process_match` verification jobs
+        /// Uses the first verification job id in the list of ids for the process_match proofs
+        /// as the mapping key.
+        process_match_callback_elems: LegacyMap<felt252,
+        StorageAccessSerdeWrapper<ProcessMatchCallbackElems>>,
```
joeykraut: Can we rename this to be a bit more clear what's going on, and expand the comments with a description of how these values are reused.

Re: naming; a better name might be `new_wallet_public_inputs` or `new_wallet_state_elements`, something of that sort

src/darkpool.cairo
```diff
@@ -378,26 +425,48 @@ mod Darkpool {
...
+            // Kick off verification
+            _poll_new_wallet_inner(ref self, callback_elems, verification_job_id)
+        }
 
-            new_root
+        /// Poll the new wallet verification job, and if it verifies, insert the new wallet into the
+        /// merkle tree
+        /// Parameters:
+        /// - `wallet_blinder_share`: The public share of the wallet blinder, used for indexing
+        /// - `wallet_share_commitment`: The commitment to the new wallet's shares
```
joeykraut: out of date docstring

src/darkpool.cairo
```diff
@@ -409,35 +478,58 @@ mod Darkpool {
...
 
-            new_root
+        /// Poll the update wallet verification job, and if it verifies, insert the new wallet into the
+        /// merkle tree
+        /// Parameters:
+        /// - `wallet_blinder_share`: The public share of the wallet blinder, used for indexing
+        /// - `wallet_share_commitment`: The commitment to the updated wallet's shares
+        /// - `old_shares_nullifier`: The nullifier for the public shares of the wallet before it was updated
+        /// - `public_wallet_shares`: The public shares of the wallet after it was updated
+        /// - `external_transfers`: The external transfers (ERC20 deposit/withdrawal)
```
joeykraut: same as above, let's just have a quick pass over the docstrings and make sure they match the interfaces after the update

src/darkpool.cairo
```diff
@@ -456,22 +550,76 @@ mod Darkpool {
             match_witness_commitments: Array<EcPoint>,
             settle_proof: Proof,
             settle_witness_commitments: Array<EcPoint>,
-        ) -> Scalar {
-            // TODO: Add verification of `VALID_MATCH_MPC` and `VALID_SETTLE`
-
-            let nullifier_set = _get_nullifier_set(@self);
-            nullifier_set.mark_nullifier_used(party_0_payload.old_shares_nullifier);
-            nullifier_set.mark_nullifier_used(party_1_payload.old_shares_nullifier);
+            verification_job_ids: Array<felt252>,
```
joeykraut, line 547: do we check anywhere that these are unique wrt other contract calls? If not we should, this seems especially important given that for this call only the first ID is used to index state

joeykraut (approved):

src/darkpool.cairo
akirillo: Haha, i actually started with `public_inputs` but decided against it since it's not *all* of the public inputs. I kind of like `callback_elems` since it feels most to the point - these are used in the "callback" logic after polling is complete. Will update comments to use this verbage as well.

src/darkpool.cairo
akirillo: In the verifier contract(s), we assure that each new `verification_job_id` is unique. So in the initial enqueuing of the verification job in `new_wallet`, `update_wallet`, `process_match`, the tx would revert if the verification job(s) were duplicates. In subsequent calls to the `poll_*` functions, it feels fair to put the responsibility onto the user to remember & reuse the correct verification job ids (not a security violation, may just drive forward someone else's verification / get a reverted tx)

src/darkpool.cairo
joeykraut: I'd prefer state elements in that case, it's unclear what "callback" means to a third party

PR #59, opened 2023-08-14
tests: darkpool: port last modified & nullifier tests

akirillo:
This PR ports over tests checking that `get_wallet_blinder_tx` and `is_nullifier_used` behave as expected in the darkpool, accounting for asynchronous verification (i.e. the `last_modified` and `nullifier_set` storage variables are updated accordingly).

**Testing**:
The new tests pass.


tests/tests/darkpool.rs
```diff
@@ -111,3 +115,187 @@ async fn test_process_match_root() -> Result<()> {
...
+        args.verification_job_id,
+    )
+    .await?
+    .is_none()
+    {
+        tx_hash = poll_new_wallet(&account, args.verification_job_id).await?;
+    }
+
+    let last_modified_tx =
+        get_wallet_blinder_transaction(&account, args.wallet_blinder_share).await?;
```
joeykraut, line 142: Now that I'm seeing these test cases, we probably want the `last_modified_tx` to actually be the first tx, not the last one. This makes indexing easy because we can just pull the public shares directly out of the tx calldata, and not have to work backwards.

For other state elements -- nullifiers, commitments, etc -- these should still be set only after verification completes.

joeykraut (approved):
Approving because we can make the change I suggested in a follow-up

tests/tests/darkpool.rs
akirillo: ack

PR #60, opened 2023-08-15
tests: darkpool: ported deposit/withdrawal tests

akirillo:
This PR ports the deposit / withdrawal tests from Cairo to the e2e test stack to account for asynchronous verification.

**Testing**:
New tests pass, other tests unaffected


tests/src/darkpool/utils.rs
```diff
@@ -146,6 +209,25 @@ pub async fn poll_new_wallet(
     .map(|r| r.transaction_hash)
 }
 
+pub async fn new_wallet_and_poll(
```
joeykraut: Maybe `poll_new_wallet_to_completion` or similar, `new_wallet_and_poll` gives little description of the end state

joeykraut: and then same renaming suggestion elsewhere this pattern is used

tests/src/darkpool/utils.rs
```diff
@@ -226,6 +327,58 @@ pub async fn process_match_verification_jobs_are_done(
...
+        // Grab the least significant 128 bits for the u128
+        low_cursor.set_position(16);
+        let low = low_cursor.read_u128::<BigEndian>().unwrap();
+
+        let mut high_cursor = Cursor::new(r[1].to_bytes_be());
+        // Grab the least significant 128 bits for the u128
+        high_cursor.set_position(16);
+        let high = high_cursor.read_u128::<BigEndian>().unwrap();
+
+        StarknetU256 { low, high }
```
joeykraut: This code is hard to read, a few suggestions, some of 
 - Add constants for `N_BYTES_FELT` and `N_BYTES_U128` and change the cursor seek to `Cursor::set_position(N_BYTES_FELT - N_BYTES_U128`)
 - Consider breaking the re-used logic out into a helper that self documents what is going on, i.e. `parse_128_least_sig_bits` or similar

joeykraut (approved):
Some readability suggestions

tests/src/darkpool/utils.rs
akirillo: ya i like this

PR #63, opened 2023-08-16
renegade_contracts, tests: align statement types w/ relayer for `VALID WALLET CREATE` & `VALID WALLET UPDATE`

akirillo:
This PR ports over statement types for `VALID WALLET CREATE` & `VALID WALLET UPDATE` from the relayer to the contracts, and directly uses relayer statement types in the e2e tests (!!).

All `new_wallet` and `update_wallet`-invoking tests pass.

Once https://github.com/renegade-fi/renegade/pull/170 is merged I'll run another `cargo update` and push that here


src/darkpool.cairo
```diff
@@ -690,61 +685,46 @@ mod Darkpool {
...
-            // It's safe to unwrap here since we only enter this branch if
-            // the list is not empty
-            let next_transfer = transfers.pop_front().unwrap();
+        // Get contract dispatcher instance for transfer mint
+        let erc20 = _get_erc20(transfer.mint);
 
-            // Get contract dispatcher instance for transfer mint
-            let erc20 = _get_erc20(next_transfer.mint);
+        // Execute the transfer
+        if !transfer.is_withdrawal {
```
joeykraut: nit: reads slightly cleaner if we switch the order of if and else, i.e. "if `is_withdrawal` do withdrawal else do deposit" instead of what we currently have which is "if not `is_withdrawal` do deposit else do withdrawal`:
```rust
if transfer.is_withdrawal {
    // Execute withdrawal
} else {
    // Execute deposit
}
```

src/darkpool/types.cairo
```diff
@@ -55,3 +76,37 @@ struct ProcessMatchCallbackElems {
     party_1_old_shares_nullifier: Scalar,
     tx_hash: felt252,
 }
+
+// -------------------
+// | STATEMENT TYPES |
+// -------------------
+/// All statement types are assumed to have the same serialization
```
joeykraut: We may want to separate out these types into separate files for discoverability

tests/Cargo.toml
```diff
@@ -25,7 +25,10 @@ byteorder = "1.4.3"
...
 katana-core = { git = "https://github.com/dojoengine/dojo.git", rev = "e305d4d" }
 dojo-test-utils = { git = "https://github.com/dojoengine/dojo.git", rev = "e305d4d" }
-mpc-stark = { git = "https://github.com/renegade-fi/mpc-stark.git" }
-mpc-bulletproof = { git = "https://github.com/renegade-fi/mpc-bulletproof.git", branch = "stark-curve-refactor", features = ["integration_test"] }
+mpc-stark = { workspace = true }
+mpc-bulletproof = { git = "https://github.com/renegade-fi/mpc-bulletproof.git", features = ["integration_test"] }
 merlin = { git = "https://github.com/renegade-fi/merlin.git" }
-renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/crate-refactor" }
+renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git" }
+circuits = { git = "https://github.com/renegade-fi/renegade.git", features = ["test_helpers"] }
```
joeykraut, line 32: can we only activate the `"test_helpers"` feature flag; e.g:
- If we only use `circuits` for testing, places this under `[dev-dependencies]`
- If we use `circuits` elsewhere as well, use:
```toml
[[test]]
required-features = ["circuits/test_helpers"]
```

I don't like the idea of having testing code in production scope, especially testing code that exposes incorrectly sized circuits.

joeykraut (approved):
Would like the dependency comment to be fixed before merging, but approving optimistically

tests/Cargo.toml
akirillo: these are arguably all dev dependencies since it's for the `tests` crate, cool if i just move everything to `[dev-dependencies]`?

akirillo: ok, have poked around with a couple options, and don't really like either:

1. moving all of the contents of `tests/src` into a module inside of `tests/tests` so that all of the dependencies can be listed under `dev-dependencies`
2. specifying `required-features` for the `[[test]]` target of the `darkpool` integration test

option 1 was a non-starter for two reasons:
1. `rust-analyzer` and `rustc` throw a huge number of unused code warnings because utilities that are not used in the compiled integration test are considered unused. This is not only annoying but makes it hard to determine what warnings are legitimate during development.
3. (this was a long and annoying rabbit hole that I could find no solution to) log output from the `tracing` crate **completely disappears**. `--nocapture` doesn't work, using `.with_test_writer()` when building the tracing subscriber doesn't work, using the `test_log` crate doesn't work.

option 2 works, but isn't great because it causes `rust-analyzer` to throw a bunch of import errors for imports from the `test_helpers` modules. this isn't the end of the world, but can make it easy to overlook other import errors during development.

to these ends, is the way this is set up now really an issue? I guess I'm not sure what you mean by "production scope". this package (`tests`) is pretty explicitly just for testing, and the `Cargo.toml` in question is for the `tests` package, not `starknet-scripts`, which I would consider "production scope." `tests` is also explicitly not a default member of the workspace.

thoughts?

joeykraut: Are the import errors "module not found"? If so you can add `"rust-analyzer.cargo.features": "[your_test_feature]",` or `"rust-analyzer.cargo.features": "all"` to your `.vscode/settings.json` file

akirillo: talked in person, for transparency: since this only affects the `tests` crate, will keep the dependencies the same

PR #64, opened 2023-08-16
renegade_contracts, tests: align `process_match` statement types

akirillo:
This PR introduces statement types for the `VALID_COMMITMENTS`, `VALID_REBLIND`, `VALID_SETTLE`, & `VALID_MATCH_MPC` circuits into the darkpool contract, and updates the e2e `process_match` tests to make use of the associated statement types from the relayer repo.

All updated tests pass.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -18,7 +18,7 @@ use types::{
     ExternalTransfer, MatchPayload, NewWalletCallbackElems, UpdateWalletCallbackElems,
     ProcessMatchCallbackElems,
 };
-use statements::{ValidWalletCreateStatement, ValidWalletUpdateStatement};
+use statements::{ValidWalletCreateStatement, ValidWalletUpdateStatement, ValidSettleStatement};
 
 
 #[starknet::interface]
@@ -55,8 +55,8 @@ trait IDarkpool<TContractState> {
         ref self: TContractState,
         wallet_blinder_share: Scalar,
         statement: ValidWalletCreateStatement,
-        proof: Proof,
         witness_commitments: Array<EcPoint>,
+        proof: Proof,
         verification_job_id: felt252,
     ) -> Option<Result<Scalar, felt252>>;
     fn poll_new_wallet(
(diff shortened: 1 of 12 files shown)


src/darkpool.cairo
```diff
@@ -77,10 +77,11 @@ trait IDarkpool<TContractState> {
         ref self: TContractState,
         party_0_payload: MatchPayload,
         party_1_payload: MatchPayload,
-        match_proof: Proof,
-        match_witness_commitments: Array<EcPoint>,
-        settle_proof: Proof,
-        settle_witness_commitments: Array<EcPoint>,
+        valid_match_mpc_witness_commitments: Array<EcPoint>,
```
joeykraut, line 80: wdyt about structuring this calldata the same way we have the statement. I'm not that opinionated, because we aren't "generating constraints" verifier side in a way that necessitates interpretable data structures, but wanted to bring this up for discussion

src/darkpool/statements.cairo
```diff
@@ -37,3 +37,48 @@ struct ValidWalletUpdateStatement {
...
+    /// A commitment to the private secret shares of the reblinded wallet
+    reblinded_private_shares_commitment: Scalar,
+    /// The global merkle root to prove inclusion into
+    merkle_root: Scalar,
+}
+
+/// Statememt for the VALID_COMMITMENTS proof
+#[derive(Drop, Serde, Copy)]
+struct ValidCommitmentsStatement {
+    /// The index of the balance that holds the mint that the wallet will
```
joeykraut: nit: phrasing is a bit odd here, I know this is copied from my code haha, but could we change it to something like:
```
The index of the balance sent by the party if a successful match occurs
```
Or something like that. Feel free to rephrase other comments as well

tests/src/utils.rs
```diff
@@ -446,58 +460,138 @@ impl CalldataSerializable for R1CSProof {
...
 }
 
-impl CalldataSerializable for MatchPayload {
+impl<const MAX_BALANCES: usize, const MAX_ORDERS: usize, const MAX_FEES: usize> CalldataSerializable
+    for ValidWalletUpdateStatement<MAX_BALANCES, MAX_ORDERS, MAX_FEES>
+where
+    [(); MAX_BALANCES + MAX_ORDERS + MAX_FEES]: Sized,
+{
+    fn to_calldata(&self) -> Vec<FieldElement> {
+        // Matches serialization expected by derived Serde impl in Cairo
```
joeykraut, line 500: Let's add a test to check that this invariant holds, I can see a world in which one side of this integration changes and is left uncaught

joeykraut (approved):

src/darkpool.cairo
akirillo: yeah, we never need to access the witness commitments in a structured way, and we'd be eating serialization cost if we did this, so i'm pretty strongly opposed

PR #65, opened 2023-08-17
renegade_contracts: add `to_scalars` serialization for statement types

akirillo:
This PR introduces a `ScalarSerializable` trait which allows a type to be serialized into an `Array<Scalar>`, and implements this trait for each of the statement types in a way that is consistent with the relayer-side implementation of the `to_scalars` method on the `BaseType` trait. This is in preparation for witness injection.

New tests asserting that the `to_scalars` serializations match have been added and pass.

### DIFF
--- a/src/darkpool/types.cairo
+++ b/src/darkpool/types.cairo
@@ -1,9 +1,12 @@
-use traits::TryInto;
+use traits::{TryInto, Into};
 use option::OptionTrait;
 use clone::Clone;
+use array::ArrayTrait;
 use starknet::ContractAddress;
 
-use renegade_contracts::{verifier::{scalar::Scalar, types::Proof}, utils::serde::EcPointSerde};
+use renegade_contracts::{
+    verifier::{scalar::{Scalar, ScalarSerializable}, types::Proof}, utils::serde::EcPointSerde
+};
 
 use super::statements::{ValidReblindStatement, ValidCommitmentsStatement};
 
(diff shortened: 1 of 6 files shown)


src/darkpool/types.cairo
```diff
@@ -35,6 +38,23 @@ impl ExternalTransferDefault of Default<ExternalTransfer> {
     }
 }
 
+impl ExternalTransferToScalarsImpl of ScalarSerializable<ExternalTransfer> {
+    fn to_scalars(self: @ExternalTransfer) -> Array<Scalar> {
+        let mut scalars: Array<Scalar> = ArrayTrait::new();
+
+        scalars.append((*self.account_addr).into());
```
joeykraut, line 45: what happens if the account addr doesn't fit in the `Scalar` field? same with the mint

joeykraut (approved):

src/darkpool/types.cairo
akirillo: hmm, good point. the way `Into` is implemented, they'll get reduced modulo the scalar field modulus. how is this implemented relayer-side? I'm thinking both contract and relayer-side, we should convert to a Starknet `u256`, i.e. serialize as 2 scalars. thoughts?

akirillo: looks like relayer-side, `to_scalars` for a `BigUint` is a `vec![biguint_to_scalar(big_uint)]`, which just calls out to `Scalar::from(big_uint)`, which calls out to the underlying arkworks `Fp::from(big_uint)`, which also reduces mod the field order.

akirillo: per our async discussion: tracked in our internal bug tracker, will be handled later

PR #66, opened 2023-08-17
renegade_contracts: darkpool: inject statement into witness

akirillo:
This PR introduces witness injection, i.e. converts the proof statements into an array of scalars, computes a pedersen commitment to each, and appends these commitments to the witness commitments. This ordering should be consistent with what is in the relayer.

**Testing**
The current darkpool tests will fail in response to these changes. Now that we are injecting (real) statements into (dummy) witnesses, the dummy circuit proof witnesses effectively become malformed and will not verify.
This PR deliberately doesn't make any testing changes yet, as I need to weight out the options of either 1) forking the darkpool contract & making dummy statement types just for testing, or 2) configuring the e2e tests to use our actual circuits (!!)

### DIFF
--- a/src/utils/crypto.cairo
+++ b/src/utils/crypto.cairo
@@ -1,0 +1,76 @@
+use traits::{TryInto, Into};
+use option::OptionTrait;
+use array::{ArrayTrait, SpanTrait};
+use keccak::keccak_u256s_le_inputs;
+use ec::ec_mul;
+
+use alexandria::data_structures::array_ext::ArrayTraitExt;
+use renegade_contracts::verifier::scalar::{Scalar, ScalarSerializable};
+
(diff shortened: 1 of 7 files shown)


src/utils/crypto.cairo
```diff
@@ -0,0 +1,76 @@
...
+            },
+        };
+    };
+
+    commitments
+}
+
+fn inject_witness<T, impl TScalarSerializable: ScalarSerializable<T>>(
+    B: EcPoint, B_blind: EcPoint, statement: @T, ref witness_commitments: Array<EcPoint>
+) {
```
joeykraut: 1. Let's add a docstring
2. I'd prefer a different name; maybe `append_statement_commitments`. This method doesn't really "inject" a witness into anything and I believe the calling code reads quite differently with the name `inject_witness`

joeykraut (approved):
Clean, one suggestion on naming

PR #67, opened 2023-08-18
renegade_contracts: darkpool: multiple verifier contracts

akirillo:
This PR introduces usage of multiple verifier contracts in the darkpool for each separate circuit.

New Cairo tests have been added for upgrading the verifier contracts and ensuring that upgrading is access-gated.

E2E tests for this (and witness injection) are left for a future PR, I'l be creating a new dummy circuit that expects the actual circuits' statements but does nothing with them, keeping the circuit small but allowing us to accurately test both multi-circuit verification and witness injection.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -320,12 +320,11 @@ mod Darkpool {
             circuit_params: CircuitParams
         ) {
             ownable__assert_only_owner(@self);
-            let verifier = _get_verifier(@self, circuit);
 
             // Save verifier contract address to storage
             _write_verifier_address(ref self, circuit, verifier_contract_address);
             // Initialize the verifier
-            verifier.initialize(circuit_params);
+            _get_verifier(@self, circuit).initialize(circuit_params);
         }
 
         // -----------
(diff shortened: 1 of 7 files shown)


src/darkpool.cairo
```diff
@@ -710,11 +755,22 @@ mod Darkpool {
...
-        IVerifierDispatcher { contract_address: self.verifier_contract_address.read() }
+    fn _get_verifier(self: @ContractState, circuit: Circuit) -> IVerifierDispatcher {
+        let contract_address = match circuit {
+            Circuit::ValidWalletCreate(()) => self.valid_wallet_create_verifier_address.read(),
+            Circuit::ValidWalletUpdate(()) => self.valid_wallet_update_verifier_address.read(),
+            Circuit::ValidCommitments(()) => self.valid_commitments_verifier_address.read(),
+            Circuit::ValidReblind(()) => self.valid_reblind_verifier_address.read(),
+            Circuit::ValidMatchMpc(()) => self.valid_match_mpc_verifier_address.read(),
+            Circuit::ValidSettle(()) => self.valid_settle_verifier_address.read(),
+        };
```
joeykraut: Just call `_read_verifier_address`?

joeykraut (approved):

src/darkpool.cairo
akirillo: good catch, added that later

PR #68, opened 2023-08-19
tests: darkpool: testing multi-circuit verification

akirillo:
This PR refactors the `darkpool` e2e tests and utilities to be aware of multi-circuit verification.

Primarily, this means we are now deploying & initializing separate verifier contracts, and using new dummy circuits which expect the right witness type (so that contract-side appending of the statement into the witness works properly), but do almost nothing (so that the circuit itself stays small).

The updated tests (and those affected by changes to the utilities) pass.

This PR does NOT make all the necessary changes to the deploy scripts so that they properly deploy and initialize all of the verifier contracts, this is low priority at the moment and will be left for a future PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ starknet = "0.5.0"
 mpc-stark = "0.2"
 
 [patch."https://github.com/starkware-libs/blockifier"]
-blockifier = { git = "https://github.com/dojoengine/blockifier", rev = "c794d1b" }
+blockifier = { git = "https://github.com/renegade-fi/blockifier.git", rev = "9c9bbe1" }
 
 [patch.crates-io]
 cairo-felt = { git = "https://github.com/dojoengine/cairo-rs.git", rev = "262b7eb4b11ab165a2a936a5f914e78aa732d4a2" }
@@ -34,3 +34,4 @@ cairo-lang-sierra-to-casm = { git = "https://github.com/starkware-libs/cairo.git
 cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo.git", tag = "v2.1.1" }
 cairo-lang-syntax = { git = "https://github.com/starkware-libs/cairo.git", tag = "v2.1.1" }
 cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo.git", tag = "v2.1.1" }
+cairo-lang-sierra-type-size = { git = "https://github.com/starkware-libs/cairo.git", tag = "v2.1.1" }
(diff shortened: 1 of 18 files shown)


tests/src/verifier/utils.rs
```diff
@@ -98,3 +133,146 @@ pub async fn step_verification(
     .await
     .map(|_| ())
 }
+
+// -----------------
```
akirillo, line 150: this was moved from `tests/src/utils.rs` since it is now only used for the `verifier` e2e tests (no longer the `darkpool` ones)

Cargo.toml
```diff
@@ -5,10 +5,10 @@ resolver = "2"
 
 [workspace.dependencies]
 starknet = "0.5.0"
-mpc-stark = "0.2"
+mpc-stark = "0.2.2"
```
joeykraut: Let's leave it at `0.2` and just run `cargo update`, this should get the most recent patch version every time, so we don't have to continue bumping it

src/darkpool.cairo
```diff
@@ -954,51 +940,55 @@ mod Darkpool {
...
 
-        // Step through VALID SETTLE
-        let valid_settle_verifier = _get_verifier(@self, Circuit::ValidSettle(()));
-        let valid_settle_verified = valid_settle_verifier
-            .step_verification(*verification_job_ids[5]);
-        match valid_settle_verified {
+                    i += 1;
+                },
+                Option::None(()) => {
+                    all_verified = Option::None(());
```
joeykraut: why do we need to reset this variable here? if it's entering into this state it should be `Option::None(())` and we should guarantee that elsewhere.

Somewhat related: do we have a check anywhere to ensure that the verification status is `Option::None` when we call `poll`? If not we should. For example I especially shouldn't be able to call poll when my verification status is `Option::Some(false)`. That implies this line should always be a no-op

src/verifier.cairo
```diff
@@ -55,7 +55,7 @@ mod Verifier {
     // | CONSTANTS |
     // -------------
 
-    const STEP_UNIT_GAS: u128 = 1000000;
+    const MSM_CHUNK_SIZE: usize = 50;
```
joeykraut, line 62: How did we come to choose 50? Can we document this so that we know how to update it when the assumptions underlying the choice change?

joeykraut (approved):

src/verifier.cairo
akirillo: honestly, fairly arbitrary for now as this is all pre-release code, this was a number that made tests go sufficiently fast and didn't force the vm out of steps. but will document this, ya

PR #70, opened 2023-08-21
starknet-scripts: Add `dump-deployments` CLI option

joeykraut:
### Purpose
This PR adds the `dump-deployments` CLI option to the `starknet-scripts`. When this flag is activated, the script will dump a JSON representation of the deployment state to `$(pwd)/deployments.json`.

This will be used by the `renegade` integration tests to pull the contract address at runtime from the dumped state in the Katana container.

### Testing
- Deployed to a local Katana node with `--dump-deployments` and verified the `deployments.json` file was created as expected

### DIFF
--- a/starknet_scripts/src/commands/utils.rs
+++ b/starknet_scripts/src/commands/utils.rs
@@ -1,4 +1,5 @@
-use eyre::Result;
+use eyre::{Context, Result};
+use json::JsonValue;
 use starknet::{
     accounts::{Account, Call, SingleOwnerAccount},
     contract::ContractFactory,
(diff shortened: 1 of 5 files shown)


starknet_scripts/src/commands/utils.rs
```diff
@@ -94,6 +99,35 @@ pub fn get_artifacts(artifacts_path: &str, contract_name: &str) -> (PathBuf, Pat
...
+        json::parse(&file_contents)?
+    } else {
+        json::object! {
+            deployments: {}
+        }
+    };
+
+    // Add a new deployment to the file contents
+    deployments["deployments"][contract.to_string().to_lowercase()] =
+        JsonValue::String(format!("0x{addr:x}"));
```
akirillo, line 122: if you've tested loading these deployments into `FieldElement`s and it works, then this is fine, but I feel like I remember running into expected hex length issues and as such use the `{:#64x}` formatter throughout my code

akirillo (approved):
lgtm, small suggestion that i honestly forget if its relevant or not

starknet_scripts/src/commands/utils.rs
joeykraut: The `starknet-client` that is consuming them loads them in (or should) in a length independent way

PR #71, opened 2023-08-22
tests: devnet state dumping/loading

akirillo:
This PR introduces post-contract-initialization state dumping / loading functionality to the e2e tests, in order to speed up dev cycles.

Each e2e test suite gets a different state dump, and all test threads for that suite will either start by loading from the dumped state (if the `LOAD_STATE` env var is set), or will block on the first thread to initialize & dump state, then proceed by loading.

All e2e test suites (except for the darkpool, which is in an unstable state right now pending #68) pass.

### DIFF
--- a/tests/src/utils.rs
+++ b/tests/src/utils.rs
@@ -51,7 +51,6 @@ use starknet_scripts::commands::utils::{calculate_contract_address, get_artifact
 use std::{
     env,
     fs::{self, File},
-    future::Future,
     io::Cursor,
     iter,
     path::{Path, PathBuf},
@@ -61,7 +60,19 @@ use tokio::sync::Mutex;
 use tracing::debug;
 use tracing_subscriber::{fmt, EnvFilter};
 
-use crate::merkle::{ark_merkle::ScalarMerkleTree, utils::TEST_MERKLE_HEIGHT};
+use crate::{
+    darkpool::utils::{init_darkpool_test_state, init_darkpool_test_statics},
+    merkle::{
+        ark_merkle::ScalarMerkleTree,
+        utils::{init_merkle_test_state, init_merkle_test_statics, TEST_MERKLE_HEIGHT},
+    },
+    nullifier_set::utils::{init_nullifier_set_test_state, init_nullifier_set_test_statics},
+    poseidon::utils::{init_poseidon_test_state, init_poseidon_test_statics},
+    statement_serde::utils::{init_statement_serde_test_state, init_statement_serde_test_statics},
+    transcript::utils::{init_transcript_test_state, init_transcript_test_statics},
+    verifier::utils::{init_verifier_test_state, init_verifier_test_statics},
+    verifier_utils::utils::{init_verifier_utils_test_state, init_verifier_utils_test_statics},
+};
 
 // ---------------------
 // | META TEST HELPERS |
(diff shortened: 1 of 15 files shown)


tests/src/utils.rs
```diff
@@ -78,26 +95,113 @@ fn get_test_starknet_config() -> StarknetConfig {
...
+    artifacts_path: &str,
+    contract_name: &str,
+    constructor_calldata: &[FieldElement],
+) -> Result<FieldElement> {
+    let class_hash = get_sierra_class_hash_from_artifact(artifacts_path, contract_name)?;
+    Ok(calculate_contract_address(class_hash, constructor_calldata))
+}
+
+pub async fn setup_sequencer<Fut: Future<Output = Result<TestSequencer>>>(
+    state_dumped_lock: &'static Mutex<bool>,
```
joeykraut: If we're making it static, why not just reference the static directly here? This simplifies the interface and allows the `setup_sequencer` to completely encapsulate the concurrency management; i.e. we do not leak or couple this with the caller

joeykraut: It looks like this is done to separate out tests, i.e. Merkle vs Darkpool; but it seems that this is overcomplicating it, why not just have a global setup and simplify the harness?

tests/src/utils.rs
```diff
@@ -78,26 +95,113 @@ fn get_test_starknet_config() -> StarknetConfig {
...
+        if sequencer.is_some() {
+            sequencer.unwrap()
+        } else {
+            let sequencer = global_setup(Some(load_state(state_separator).await?)).await;
+            debug!("Loaded state");
+            sequencer
+        }
+    };
+
+    Ok(sequencer)
```
joeykraut: This code is more complex than it needs to be, I'd suggest we simplify it as follows:
```rust
let should_load = env::var(LOAD_STATE_ENV_VAR).is_ok();
let already_initialized = state_dumped_lock.lock().await;

let sequencer = if should_load || *already_initialized {
    drop(already_initialized);
    global_setup(...)
} else {
    let sequencer = state_init_fut.await?;
    dump_state(...);
    *state_dumped = true;
    sequencer
}
```

tests/src/utils.rs
```diff
@@ -78,26 +95,113 @@ fn get_test_starknet_config() -> StarknetConfig {
...
+    contract_name: &str,
+    constructor_calldata: &[FieldElement],
+) -> Result<FieldElement> {
+    let class_hash = get_sierra_class_hash_from_artifact(artifacts_path, contract_name)?;
+    Ok(calculate_contract_address(class_hash, constructor_calldata))
+}
+
+pub async fn setup_sequencer<Fut: Future<Output = Result<TestSequencer>>>(
+    state_dumped_lock: &'static Mutex<bool>,
+    state_separator: &str,
```
joeykraut: We only use one value here, let's save complexity where we can and remove this parameter. This can then trickle down into the helpers this method calls

tests/src/utils.rs
```diff
@@ -78,26 +95,113 @@ fn get_test_starknet_config() -> StarknetConfig {
...
+}
+
+pub async fn dump_state(sequencer: &TestSequencer, separator: &str) -> Result<()> {
+    let state_path = get_state_path(separator);
+    let state = sequencer.sequencer.backend.dump_state().await?;
+    fs::write(state_path, state).map_err(|e| eyre!("Error dumping state: {e}"))
+}
+
+pub async fn load_state(
+    // sequencer: &mut TestSequencer,
```
joeykraut: dead code

joeykraut (commented):
Want to see if we can pare down the harness complexity a bit before approving. My suggestions broadly boil down to:

We over-parameterize the setup interfaces to fit each "type" of test, in the end, the setup helpers don't do much for us as the meat of the logic lives in callbacks. I'd suggest we have a single initialization flow that deploys _all_ contracts even if they aren't used for a given test. If this is too burdensome, we can have a flag structure like `SetupArgs` that decides which contracts to deploy.

This will simplify the concurrency model and allow for much more readable, simple code that applies generally as a bonus.

tests/src/utils.rs
akirillo: this is on me, we use a constant that has the same name but is defined with different values across the various test scaffolding files. will rename to make more clear.

tests/src/utils.rs
akirillo: hmm, i do like the early exit if `LOAD_STATE_ENV_VAR` is set, even though it leads to some duplicate code - in your suggestion here, we would have to acquire the lock even if the env var is set (even though we immediately release it), which we don't have to do

akirillo: hmm, i do like the early exit if `LOAD_STATE_ENV_VAR` is set, even though it leads to some duplicate code - in your suggestion here, we would have to acquire the lock even if the env var is set (even though we immediately release it), which we don't have to do

tests/src/utils.rs
```diff
@@ -78,26 +139,133 @@ fn get_test_starknet_config() -> StarknetConfig {
...
+        TestConfig::NullifierSet => (&NULLIFIER_SET_STATE_DUMPED, NULLIFIER_SET_STATE_SEPARATOR),
+        TestConfig::Verifier => (&VERIFIER_STATE_DUMPED, VERIFIER_STATE_SEPARATOR),
+        TestConfig::VerifierUtils => (&VERIFIER_UTILS_STATE_DUMPED, VERIFIER_UTILS_STATE_SEPARATOR),
+        TestConfig::Transcript => (&TRANSCRIPT_STATE_DUMPED, TRANSCRIPT_STATE_SEPARATOR),
+        TestConfig::Poseidon => (&POSEIDON_STATE_DUMPED, POSEIDON_STATE_SEPARATOR),
+        TestConfig::StatementSerde => (
+            &STATEMENT_SERDE_STATE_DUMPED,
+            STATEMENT_SERDE_STATE_SEPARATOR,
+        ),
+    };
```
joeykraut: nit: I'd break this out into a helper just for readability

joeykraut: Maybe below as well, I feel that this function is most readable when restricted only to its control flow (since this is a rather complex function), so the `match` clauses may be better off in a set of helpers

joeykraut (approved):
Looks a lot better, great job, I think from the perspective of the test logic; the setup interface is super clean!

PR #72, opened 2023-08-25
renegade_contracts: nullifier_set, darkpool: add in-progress nullifier set

akirillo:
This PR introduces an "in-progress nullifiers" set to the `NullifierSet` contract, which is used to track nullifiers that are currently in use in a verification job (preventing, for example, races of two different matches being settled simultaneously).

This PR also adds a couple basic unit tests for the new functionality, but tests in the e2e suite ensuring that this prevents a double `process_match` or `update_wallet` are left for the next PR.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -51,8 +51,7 @@ trait IDarkpool<TContractState> {
     ) -> felt252;
     fn get_root(self: @TContractState) -> Scalar;
     fn root_in_history(self: @TContractState, root: Scalar) -> bool;
-    fn is_nullifier_used(self: @TContractState, nullifier: Scalar) -> bool;
-    fn is_nullifier_in_progress(self: @TContractState, nullifier: Scalar) -> bool;
+    fn is_nullifier_available(self: @TContractState, nullifier: Scalar) -> bool;
     fn check_verification_job_status(
         self: @TContractState, circuit: Circuit, verification_job_id: felt252
     ) -> Option<bool>;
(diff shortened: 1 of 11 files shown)


src/nullifier_set.cairo
```diff
@@ -4,7 +4,10 @@ use renegade_contracts::verifier::scalar::Scalar;
 #[starknet::interface]
 trait INullifierSet<TContractState> {
     fn is_nullifier_used(self: @TContractState, nullifier: Scalar) -> bool;
+    fn is_nullifier_in_progress(self: @TContractState, nullifier: Scalar) -> bool;
     fn mark_nullifier_used(ref self: TContractState, nullifier: Scalar);
+    fn mark_nullifier_in_progress(ref self: TContractState, nullifier: Scalar);
+    fn mark_nullifier_not_in_progress(ref self: TContractState, nullifier: Scalar);
```
joeykraut: I'd prefer a different name that indicated what the state _is_ after this operation, not what the state _isn't_. Maybe `mark_nullifier_unused` or `forget_nullifier`?

src/nullifier_set.cairo
```diff
@@ -30,10 +36,16 @@ mod NullifierSet {
...
+    struct NullifierInProgress {
+        nullifier: Scalar, 
+    }
+
     #[event]
     #[derive(Drop, PartialEq, starknet::Event)]
     enum Event {
-        NullifierSpent: NullifierSpent, 
+        NullifierSpent: NullifierSpent,
+        NullifierInProgress: NullifierInProgress,
```
joeykraut: Can we add an event for a nullifier being removed?

src/darkpool.cairo
```diff
@@ -52,6 +52,7 @@ trait IDarkpool<TContractState> {
     fn get_root(self: @TContractState) -> Scalar;
     fn root_in_history(self: @TContractState, root: Scalar) -> bool;
     fn is_nullifier_used(self: @TContractState, nullifier: Scalar) -> bool;
+    fn is_nullifier_in_progress(self: @TContractState, nullifier: Scalar) -> bool;
```
joeykraut: It'd be nice to have a method `is_nullifier_unused` that returns:
```
!(is_nullifier_used(...) || is_nullifier_in_progress(...))
```
This is the most common question a client wants to answer wrt nullifiers

src/darkpool.cairo
```diff
@@ -552,6 +563,15 @@ mod Darkpool {
             proof: Proof,
             verification_job_id: felt252,
         ) {
+            // Assert that the `old_shares_nullifier` is not already spent
+            let nullifier_set = _get_nullifier_set(@self);
+            assert(
+                !nullifier_set.is_nullifier_used(statement.old_shares_nullifier),
+                'nullifier already used'
+            );
```
joeykraut: We don't need this assertion here right? The nullifier set implementation already asserts this

src/darkpool.cairo
```diff
@@ -658,6 +681,32 @@ mod Darkpool {
...
+                    ),
+                'nullifier already used'
+            );
+            assert(
+                !nullifier_set
+                    .is_nullifier_used(
+                        party_1_payload.valid_reblind_statement.original_shares_nullifier
+                    ),
+                'nullifier already used'
+            );
```
joeykraut: same here, it's cleaner if the nullifier interface gives protection against double spends so that the caller may avoid making these assertions

joeykraut (approved):
Small suggestions about the interface

PR #79, opened 2023-08-30
renegade_contracts: darkpool: verify `VALID WALLET UPDATE` statement signature

akirillo:
This PR introduces a pre-verification check of a signature over the `VALID WALLET UPDATE` statement generated from `sk_root`.

*Testing*
A new test was added asserting that an invalid signature fails. This, and all previous `update_wallet` tests (which now sign the statement) pass.


src/utils/crypto.cairo
```diff
@@ -67,10 +68,37 @@ fn commit_public(B: EcPoint, B_blind: EcPoint, mut inputs: Span<Scalar>) -> Arra
...
     let statement_scalars = statement.to_scalars();
     let mut statement_commitments = commit_public(B, B_blind, statement_scalars.span());
     witness_commitments.append_all(ref statement_commitments);
 }
+
+/// Computes the Keccak256 hash of the given statement, reducing the result into a Scalar.
+///
+/// We manually mirror the implementation of keccak::keccak_u256s_le_inputs so that we can
+/// avoid an extra loop over the input to map it into u256s.
+fn hash_statement<T, impl TScalarSerializable: ScalarSerializable<T>>(statement: @T) -> Scalar {
```
joeykraut, line 85: in this method, why do we sign the scalar serialization and not the byte serialization? it'd be ideal from the FE's perspective if this were byte-centric so that the FE does not have to implement scalar serialization (this also fits closer to how keccak is defined). Is this a Starknet limitation?

src/darkpool.cairo
```diff
@@ -570,6 +575,18 @@ mod Darkpool {
                 'invalid statement merkle root'
             );
 
+            // Assert that statement signature is valid
```
joeykraut: Can we make a note here in the comment on key rotation? I.e. that `old_pk_root` is no longer the root key of the wallet and that this method implicitly authorizes a key rotation as well?

joeykraut (approved):
couple questions

src/utils/crypto.cairo
akirillo: Mostly because it'll be more expensive & cumbersome contract-side to serialize the statement into bytes. Since I've last looked at the corelib, it seems some utilities for converting between `felt252` and a new `bytes31` type have been added, but the type seems a bit inconvenient (e.g. it is not a byte array, not sure how to add padding to it, etc.). Either way, this would add another layer of serialization logic and thus cost to the contract execution, and I am generally in favor of pushing serialization complexity onto the client / relayer instead of in the contract, where we pay for it. Even if the costs are marginal wrt other contract functions, might as well handle these things in a lower-cost (and more expressive) programming environment than Cairo.

src/darkpool.cairo
akirillo: Yes, good call

PR #80, opened 2023-08-31
renegade_contracts: verifier, darkpool: make verifier contract `Ownable`

akirillo:
This PR adds access control to the verifier, guarding the mutating functions (`queue_verification_job` and `step_verifier`) with an assertion that they are being called by the contract owner, which must be the darkpool.

Additionally, this PR changes the deployment pattern of the verifier contracts to be handled by the darkpool contract itself in its initialization. This makes our deployment logic simpler, and ensures that the verifier owner is the darkpool.

Cairo-native access control tests have been added for the verifier, which pass, as do all tests that had to be adapted to fit the new deployment pattern.

Relevant changes to the `starknet_scripts` crate are incomplete, a proper refactor of this crate is left for a future PR pending the completion of the verifier into the darkpool.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -30,13 +30,11 @@ trait IDarkpool<TContractState> {
         ref self: TContractState,
         merkle_class_hash: ClassHash,
         nullifier_set_class_hash: ClassHash,
+        verifier_class_hash: ClassHash,
         height: u8,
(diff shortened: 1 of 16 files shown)


akirillo:
moving to draft after our discussion around making library calls to a multi-verifier contract

PR #81, opened 2023-09-01
renegade_contracts: verifier, darkpool: multiverifier contract

akirillo:
This PR refactors the verifier to manage multiple circuits in a single contract, which bubbles up to the darkpool such that the verifier can be invoked via library calls as opposed to needing to deploy multiple verifier contracts.

All Cairo tests pass, but e2e tests still need to be updated to reflect these changes.

Also, I removed Pedersen generators from contract storage since we consistently use the STARK ECDSA generator everywhere

### DIFF
--- a/src/verifier.cairo
+++ b/src/verifier.cairo
@@ -12,28 +12,36 @@ use renegade_contracts::utils::serde::EcPointSerde;
 use types::{CircuitParams, Proof, VerificationJob};
 
 #[starknet::interface]
-trait IVerifier<TContractState> {
-    fn initialize(ref self: TContractState, circuit_params: CircuitParams);
+trait IMultiVerifier<TContractState> {
+    fn add_circuit(ref self: TContractState, circuit_id: felt252);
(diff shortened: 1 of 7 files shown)


src/verifier.cairo
```diff
@@ -496,7 +524,9 @@ mod Verifier {
         commitments_rem
     }
 
-    fn step_verification_inner(ref self: ContractState, ref verification_job: VerificationJob) {
+    fn step_verification_inner(
+        ref self: ContractState, circuit_id: felt252, ref verification_job: VerificationJob
```
joeykraut: wdyt about storing `Verificationjob -> CircuitId` to simplify this interface?

joeykraut (approved):

src/verifier.cairo
akirillo: hmm, my initial aversion is that it's an extra storage read, but I'll go ahead and do this - for now, I think we should optimize readability over performance, and leave perf optimizations as a separate effort

PR #84, opened 2023-09-02
renegade_contracts: darkpool, merkle: verifier & poseidon feature flags

akirillo:
This PR introduces an initial set of feature flags to be used in testing for the darkpool. Currently the feature flags toggle whether or not to invoke the verifier, and whether to use our implementation of poseidon hashing over the scalar field, or the native poseidon builtin.

Cairo tests have been updated to ingest these changes, and pass. Applying these changes to the e2e tests is left for the next PR.

### DIFF
--- a/src/darkpool.cairo
+++ b/src/darkpool.cairo
@@ -117,7 +117,7 @@ mod Darkpool {
         nullifier_set::{INullifierSetLibraryDispatcher, INullifierSetDispatcherTrait},
         utils::{
             serde::EcPointSerde, storage::StoreSerdeWrapper,
-            crypto::{append_statement_commitments, hash_statement}
+            crypto::{append_statement_commitments, hash_statement, native_poseidon_hash_scalars}
         },
         oz::erc20::{IERC20DispatcherTrait, IERC20Dispatcher},
     };
@@ -144,7 +144,7 @@ mod Darkpool {
         _initialized: bool,
         // CORE
         /// Feature flag settings
-        feature_flags: StoreSerdeWrapper<FeatureFlags>,
+        feature_flags: FeatureFlags,
         /// Stores the implementation class hash for the Merkle tree interface
         merkle_class_hash: ClassHash,
         /// Stores the implementation class hash for the nullifier set interface
@@ -249,7 +249,7 @@ mod Darkpool {
     #[constructor]
     fn constructor(ref self: ContractState, owner: ContractAddress, feature_flags: FeatureFlags) {
         _ownable_initialize(ref self, owner);
-        self.feature_flags.write(StoreSerdeWrapper { inner: feature_flags });
+        self.feature_flags.write(feature_flags);
     }
 
     // -------------
@@ -300,7 +300,7 @@ mod Darkpool {
 
             // Add all of the circuits to the multi-verifier
 
-            if self.feature_flags.read().inner.verifier {
+            if self.feature_flags.read().verifier {
                 let verifier = _get_verifier(@self);
                 let mut circuits = array![
                     Circuit::ValidWalletCreate(()),
@@ -324,14 +324,15 @@ mod Darkpool {
             }
 
             // Initialize the Merkle tree
-            _get_merkle_tree(@self).initialize(height);
+            _get_merkle_tree(@self)
+                .initialize(height, self.feature_flags.read().non_native_poseidon);
         }
 
         /// Adds a circuit to the verifier
         /// Parameters:
         /// - `circuit`: The circuit to add
         fn add_circuit(ref self: ContractState, circuit: Circuit) {
-            assert(self.feature_flags.read().inner.verifier, 'verifier disabled');
+            assert(self.feature_flags.read().verifier, 'verifier disabled');
             ownable__assert_only_owner(@self);
             _get_verifier(@self).add_circuit(circuit.into());
         }
@@ -343,7 +344,7 @@ mod Darkpool {
         fn parameterize_circuit(
             ref self: ContractState, circuit: Circuit, circuit_params: CircuitParams
         ) {
-            assert(self.feature_flags.read().inner.verifier, 'verifier disabled');
+            assert(self.feature_flags.read().verifier, 'verifier disabled');
             ownable__assert_only_owner(@self);
             // Parameterize the circuit
             _get_verifier(@self).parameterize_circuit(circuit.into(), circuit_params);
@@ -404,7 +405,7 @@ mod Darkpool {
         /// - `verifier_class_hash`: The hash of the implementation class used for verifier operations
         /// - `circuit`: The circuit for which to upgrade the verifier
         fn upgrade_verifier(ref self: ContractState, verifier_class_hash: ClassHash) {
-            assert(self.feature_flags.read().inner.verifier, 'verifier disabled');
+            assert(self.feature_flags.read().verifier, 'verifier disabled');
             ownable__assert_only_owner(@self);
 
             // Get existing class hash to emit event
@@ -474,7 +475,7 @@ mod Darkpool {
         fn check_verification_job_status(
             self: @ContractState, verification_job_id: felt252
         ) -> Option<bool> {
-            assert(self.feature_flags.read().inner.verifier, 'verifier disabled');
+            assert(self.feature_flags.read().verifier, 'verifier disabled');
             _get_verifier(self).check_verification_job_status(verification_job_id)
         }
 
@@ -497,7 +498,7 @@ mod Darkpool {
             proof: Proof,
             verification_job_id: felt252,
         ) {
-            if self.feature_flags.read().inner.verifier {
+            if self.feature_flags.read().verifier {
                 let verifier = _get_verifier(@self);
 
                 // Inject witness
@@ -534,7 +535,7 @@ mod Darkpool {
         fn poll_new_wallet(
             ref self: ContractState, verification_job_id: felt252, 
         ) -> Option<Result<Scalar, felt252>> {
-            let verified = if self.feature_flags.read().inner.verifier {
+            let verified = if self.feature_flags.read().verifier {
                 let verifier = _get_verifier(@self);
 
                 assert(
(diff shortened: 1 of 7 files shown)


src/darkpool.cairo
```diff
@@ -297,35 +300,39 @@ mod Darkpool {
...
-
-            loop {
-                match circuits.pop_front() {
-                    Option::Some(circuit) => {
-                        verifier.add_circuit(circuit.into());
-                    },
-                    Option::None(()) => {
-                        break;
-                    }
+            if self.feature_flags.read().verifier {
```
joeykraut: Why not just parameterize the verifier upon its initialization with this feature flag? If not active then the verifier can just immediately return "verified" in its `step_verification` implementation 

this would simplify the polling logic quite a bit i feel

src/darkpool.cairo
```diff
@@ -678,9 +710,16 @@ mod Darkpool {
...
+
+                        let total_shares_commitment = if self
+                            .feature_flags
+                            .read()
+                            .non_native_poseidon {
+                            *poseidon_hash(hash_input.span(), 1 // num_elements
+                            )[0]
+                        } else {
+                            native_poseidon_hash_scalars(hash_input.span())
+                        };
```
joeykraut: I'd suggest breaking this out into a helper (i.e. `compute_share_commitment_with_feature_flag` or similar that encapsulates this repeated branching

src/darkpool.cairo
```diff
@@ -869,71 +916,79 @@ mod Darkpool {
...
+
+                            i += 1;
+                        },
+                        Option::None(()) => {
+                            break;
+                        },
+                    };
                 };
+
+                all_verified
```
joeykraut: Do we really need this second loop here? The code would be a lot cleaner if we just poll in the first loop then break. we don't really need to `assert(should_poll)` either, so I don't see the need for one loop to decide whether to poll and one loop to do the polling.

In the case that all verifications are already complete, this code path can just be a no-op

src/darkpool.cairo
```diff
@@ -958,16 +1013,29 @@ mod Darkpool {
...
+                                )[0],
+                                *poseidon_hash(party_1_hash_input.span(), 1 // num_elements
+                                )[0],
+                            )
+                        } else {
+                            (
+                                native_poseidon_hash_scalars(party_0_hash_input.span()),
+                                native_poseidon_hash_scalars(party_1_hash_input.span())
+                            )
+                        };
```
joeykraut: same here wrt feature flag encapsulation above

src/darkpool/types.cairo
```diff
@@ -93,6 +93,15 @@ struct Signature {
     s: Scalar,
 }
 
+/// Represents which optional features to enable in the darkpool
+#[derive(Drop, Serde, Copy, starknet::Store)]
+struct FeatureFlags {
+    /// Whether or not to use Poseidon over the scalar field
+    non_native_poseidon: bool,
+    /// Whether or not to verify proofs
+    verifier: bool,
```
joeykraut: Let's reverse the meaning of this flag to `disable_verifier`:
1. It's much more self-documenting
2. The common (production) case is that the verifier is enabled, and feature flags should describe non-default behavior.

src/darkpool/types.cairo
```diff
@@ -93,6 +93,15 @@ struct Signature {
     s: Scalar,
 }
 
+/// Represents which optional features to enable in the darkpool
+#[derive(Drop, Serde, Copy, starknet::Store)]
+struct FeatureFlags {
+    /// Whether or not to use Poseidon over the scalar field
+    non_native_poseidon: bool,
```
joeykraut: nit: I'd rename this to `use_base_field_poseidon` (i.e. a verb and something more descriptive than just "non-native")

src/merkle.cairo
```diff
@@ -61,6 +63,8 @@ mod Merkle {
         /// at the given height. Used to set the sibling pathway when a subtree is
         /// filled.
         zeros: LegacyMap<u8, Scalar>,
+        /// Indicates whether or not to use Poseidon over the scalar field
+        non_native_poseidon: bool,
```
joeykraut: same here wrt naming, it's not clear what a `true` value of this field implies. The comment seems to suggest `true` would mean scalar field Poseidon, which is non-native wrt starknet but not our proof system, so it's a bit confusing

src/testing/tests/darkpool_tests.cairo
```diff
@@ -158,7 +160,7 @@ fn test_upgrade_darkpool_access() {
 fn test_upgrade_merkle_access() {
     let test_caller = contract_address_try_from_felt252(TEST_CALLER).unwrap();
     set_contract_address(test_caller);
-    let mut darkpool = setup_darkpool();
+    let mut darkpool = setup_darkpool(FeatureFlags { verifier: false, non_native_poseidon: false });
```
joeykraut: Refactor the `setup_darkpool` method to have a default interface and possible a `setup_darkpool_with_flags` interface that it calls out to. Esp since these are all the same parameterization

joeykraut (commented):
few things that would help readability

src/darkpool.cairo
akirillo: The first loop was previously added when we decided that there should be assertions that the polling isn't already complete, similar (albeit simpler) code is present in `poll_new_wallet` and `poll_update_wallet` as well. I don't mind switching to a no-op instead of an assertion, just a matter of UX, really

src/darkpool.cairo
```diff
@@ -245,8 +249,9 @@ mod Darkpool {
     // ---------------
 
     #[constructor]
-    fn constructor(ref self: ContractState, owner: ContractAddress) {
+    fn constructor(ref self: ContractState, owner: ContractAddress, feature_flags: FeatureFlags) {
```
joeykraut, line 250: You may have seen my comment in the other PR, It'd be nice to have a default constructor that sets default feature flags to avoid any sort of safety concerns in production. Does not have to be this PR

src/darkpool.cairo
```diff
@@ -318,8 +323,13 @@ mod Darkpool {
...
 
-            // Initialize the Merkle tree & verifier
-            _get_merkle_tree(@self).initialize(height);
+            // Initialize the Merkle tree
+            _get_merkle_tree(@self)
+                .initialize(height, self.feature_flags.read().use_base_field_poseidon);
+
+            // Set verification enabled flag in the verifier
+            _get_verifier(@self)
+                .set_verification_disabled(self.feature_flags.read().disable_verification);
```
joeykraut: A more extensible pattern would be to pass the `FeatureFlags` to the verifier under a `set_feature_flags` method, this way as we add more feature flags we do less re-structuring

src/darkpool.cairo
```diff
@@ -867,81 +876,74 @@ mod Darkpool {
...
-            ]
-                .span();
-
-            let verification_job_ids = verification_job_ids.span();
-
-            assert(circuits.len() == verification_job_ids.len(), 'wrong # of verification job ids');
+            ];
 
-            // Assert that no verification jobs have failed, and that not all are complete
+            let mut already_polled = false;
```
joeykraut: nit: rename to `already_completed`, we poll many times per verification job, it's a bit confusing on first read when named `already_polled`

src/darkpool.cairo
```diff
@@ -867,81 +876,74 @@ mod Darkpool {
...
             loop {
-                if i == verification_job_ids.len() {
+                if i == circuits.len() {
+                    // We only make it to this branch if the last verification job had just polled successfully
+                    polling_result = Option::Some(true);
                     break;
                 }
 
-                let verified = verifier.check_verification_job_status(*verification_job_ids[i]);
+                let already_verified = verifier.check_verification_job_status(verification_job_id);
```
joeykraut: Above, you assigned each proof a separate `verification_job_id`, i.e. `id, id + 1, ...`. I don't see where this is incremented. Won't this just always evaluate to `Some(true)` after the first proof is verified? In which case, after the first proof is verified we will only ever hit the first branch of the `match` below and never verify the other proofs?

src/darkpool.cairo
```diff
@@ -867,81 +876,74 @@ mod Darkpool {
...
+                                    polling_result = Option::Some(false);
+                                    break;
+                                }
+                            // Continue to next iteration if verification job just succeeded.
+                            // This means that we'll do another `step_verification` call in this
+                            // transaction, which may exceed the gas limit...
+                            },
+                            Option::None(()) => {
+                                break;
+                            },
```
joeykraut: Not sure if Cairo supports constant pattern matching, but if it works like rust, this can be rewritten as:
```rust
match verified {
    Option::Some(false /* success */) => {
        polling_result = Option::Some(false);
        break;
    },
    _ => { /* fallthrough to next loop */ }
}
```
would improve readability, not a fan of these multiple nested `loop` + `match` clauses. If we can bias this code path in any way towards making it easier to reason through the different control flow paths, that'd greatly help

src/darkpool.cairo
```diff
@@ -867,81 +876,74 @@ mod Darkpool {
...
-                    },
-                    Option::None(()) => {
-                        break;
-                    },
-                };
-            };
+            assert(!already_polled, 'polling already complete');
 
-            match all_verified {
+            match polling_result {
```
joeykraut: can we break this out into a helper like `handle_verfification_result` or similar? There is a lot of control flow happening in this method

joeykraut (approved):
some feedback on excess control flow hurting readability, otherwise looks good

src/darkpool.cairo
akirillo: yup i followed up on that PR, we can keep the discussion there

src/darkpool.cairo
akirillo: ah yup good catch, need to do `verification_job_id + i.into()` here

src/darkpool.cairo
akirillo: i like this, will do it for the merkle contract as well for consistency

src/darkpool.cairo
akirillo: hmm, for one Cairo doesn't seem to support constant pattern matching, but also the cases for `Option::some(true)` and `Option::None(())` are different - the former continues to the next loop, the latter breaks out of it.

agree that it's not the most readable control flow, but each of the 4 branches here have different handling logic.

akirillo: I can at least break out an extra helper to aid in readability

PR #85, opened 2023-09-03
tests: merkle, darkpool: enable feature flags in e2e tests

akirillo:
This PR updates the merkle and darkpool e2e test suites to properly enable the new feature flags.

Tests pass, as do ad-hoc tests asserting that the contracts behave as expected when the flags are disabled.

### DIFF
--- a/starknet_scripts/src/commands/utils.rs
+++ b/starknet_scripts/src/commands/utils.rs
@@ -277,7 +277,11 @@ pub async fn deploy_darkpool(
 
     // Deploy darkpool
     debug!("Deploying darkpool contract...");
-    let calldata = vec![account.address()];
+    let calldata = vec![
+        account.address(),
+        FieldElement::ZERO, /* serialization of boolean `false` indicating that poseidon over the base field should not be used */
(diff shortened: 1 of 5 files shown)


starknet_scripts/src/commands/utils.rs
```diff
@@ -277,7 +277,11 @@ pub async fn deploy_darkpool(
 
     // Deploy darkpool
     debug!("Deploying darkpool contract...");
-    let calldata = vec![account.address()];
+    let calldata = vec![
```
joeykraut, line 280: maybe we should have a constructor for the contract that sets the feature flags to their default values. Want to make absolutely sure we don't have bugs initializing our contract

joeykraut (approved):

starknet_scripts/src/commands/utils.rs
akirillo: I can certainly make a separate function in the deploy scripts that does this. at the contract level, though, there can only be one constructor, and I want it to accept the feature flags as an argument so that we can later redeploy the darkpool with different feature flags successfully (it will be deployed to a different address since it has different constructor calldata). FWIW when we go to mainnet the feature flags will be removed completely.

PR #91, opened 2023-10-09
transcript: basic implementation

akirillo:
This PR introduces a basic transcript implementation similar to the hash-chain-based transcript we were using previously.

It currently exposes methods for appending arbitrary byte-encoded messages, scalar field elements, and $\mathbb{G}_1$ curve points to the transcript, as well as squeezing challenge scalars from it.

As the verifier gets built out and the transcript gets used, parts of this interface may change, e.g. higher-level methods composed from these or different types to ingest into the transcript state.

This makes use of the `stylus_sdk::keccak` function, which invokes the VM-accelerated (and cheaply-priced) Keccak implementation.

Currently no tests are implemented, as there is not exactly a reference implementation of this transcript to test against and I want to prioritize getting a first-pass of the verifier done.

### DIFF
--- a/src/transcript/mod.rs
+++ b/src/transcript/mod.rs
@@ -1,0 +1,61 @@
+//! A simple hash-chain based transcript used for computing challenge values via the Fiat-Shamir transformation.
+
+mod errors;
+
+extern crate alloc;
+
+use alloc::vec::Vec;
+use ark_ff::PrimeField;
+use ark_serialize::CanonicalSerialize;
+use core::result::Result;
+use stylus_sdk::{alloy_primitives::B256, crypto::keccak};
+
+use crate::types::{G1Affine, ScalarField};
+
+use self::errors::TranscriptError;
+
+pub struct Transcript {
+    state: B256,
+}
+
+impl Transcript {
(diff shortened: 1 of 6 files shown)


src/transcript/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        point.serialize_compressed(&mut message)?;
+        self.append_message(label, &message);
+        Ok(())
+    }
+
+    pub fn challenge_scalar(&mut self) -> ScalarField {
+        let new_state = keccak(self.state);
+
+        let challenge =
+            ScalarField::from_le_bytes_mod_order(&[&self.state[..], &new_state[..]].concat());
```
joeykraut: Why do we use the `new_state` here as byte input? This would imply that if I squeezed two challenges in a row they are not I.I.D, in fact the high bytes of the first will equal the low bytes of the second.

joeykraut (approved):

src/transcript/mod.rs
akirillo: great catch - tried to simplify the way scalars are drawn in the jellyfish implementation, missed an initial rehashing of the current state.

PR #92, opened 2023-10-09
verifier: define `VerificationKey` struct

akirillo:
This PR introduces the `VerificationKey` struct, which is comprised of all of the pre-processed, circuit-specific information needed by the verifier to verify a proof.

I didn't include a struct representing the SRS, because the assumption is that all of the circuit preprocessing using the SRS will be done off-chain, and we'll just initialize the verifier with `VerificationKey`s for any circuits it verifies.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,0 +1,17 @@
+//! Constants that parameterize the Plonk proof system
+
+/// The number of different wire types that can be used in the gates of the circuit.
+/// Currently, this value is 3, corresponding to the following wire types:
+/// - Left input wires
+/// - Right input wires
+/// - Output wires
+pub const NUM_WIRE_TYPES: usize = 3;
+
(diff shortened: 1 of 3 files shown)


src/constants.rs
```diff
@@ -0,0 +1,17 @@
+//! Constants that parameterize the Plonk proof system
+
+/// The number of different wire types that can be used in the gates of the circuit.
+/// Currently, this value is 3, corresponding to the following wire types:
+/// - Left input wires
+/// - Right input wires
+/// - Output wires
+pub const NUM_WIRE_TYPES: usize = 3;
```
joeykraut, line 8: Is the plan to do custom gates later?

joeykraut (approved):

src/constants.rs
akirillo: yes, custom gates will be added later, at which point these constants will be updated

PR #94, opened 2023-10-10
transcript: implement Plonk transcript protocol

akirillo:
This PR implements the Plonk transcript protocol, i.e. a single method `compute_challenges` which computes all the challenge values used throughout Plonk for a given circuit + proof.

The ordering of elements absorbed / squeezed is taken from https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/proof_system/verifier.rs#L260

### DIFF
--- a/src/transcript/mod.rs
+++ b/src/transcript/mod.rs
@@ -4,13 +4,13 @@ mod errors;
 
 extern crate alloc;
 
-use alloc::vec::Vec;
+use alloc::{format, vec::Vec};
 use ark_ff::PrimeField;
 use ark_serialize::CanonicalSerialize;
 use core::result::Result;
 use stylus_sdk::{alloy_primitives::B256, crypto::keccak};
 
-use crate::types::{G1Affine, ScalarField};
+use crate::types::{Challenges, G1Affine, Proof, ScalarField, VerificationKey};
 
 use self::errors::TranscriptError;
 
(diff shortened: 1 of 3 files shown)


src/transcript/mod.rs
```diff
@@ -28,28 +28,19 @@ impl Transcript {
         self.state = keccak([&self.state[..], label, message].concat());
     }
 
-    pub fn append_scalar(
+    pub fn append_serializable<T: CanonicalSerialize>(
```
joeykraut, line 41: nice

src/transcript/mod.rs
```diff
@@ -60,3 +51,68 @@ impl Transcript {
         challenge
     }
 }
+
+/// Computes all the challenges used in the Plonk protocol,
+/// given a verification key, a proof, and a set of public inputs.
+#[inline]
```
joeykraut: why do we inline this? It's quite a large method to be forcing the compiler to inline. Generally speaking the compiler makes good choices around when to inline, and we only force it when it's a *basically* constant function. If for performance reasons, I'd hesitate to do this speculatively

src/transcript/mod.rs
```diff
@@ -60,3 +51,68 @@ impl Transcript {
...
+pub fn compute_challenges(
```
joeykraut, line 67: Can we test this method somehow against a known implementation, say jellyfish? It may be too hard, but could be nice. Doesn't have to be in this PR if so, just want to open a discussion

joeykraut (approved):

src/transcript/mod.rs
akirillo: I see, appreciate the context - the short answer here is that we inline because Jellyfish inlines their `compute_challenges`, but am happy to remove the attribute

PR #95, opened 2023-10-11
verifier, types: initialize Verifier smart contract

akirillo:
This PR defines a smart contract for the verifier, including a Solidity ABI-compatible type for the `VerificationKey` that can be kept in contract storage.

### DIFF
--- a/src/types/errors.rs
+++ b/src/types/errors.rs
@@ -1,0 +1,23 @@
+//! Errors stemming from conversions between types used throughout the system.
+
+use stylus_sdk::alloy_primitives::ruint::FromUintError;
+
+/// Errors stemming from working with Solidity-ABI-compatible storage types
+pub enum StorageError {
+    /// An error that occurred while converting between a Solidity-ABI-compatible storage type and a Rust type
+    TypeConversion,
+    /// An error that occurred while serializing Rust types before conversion, or deserializing after conversion
+    Serialization,
+}
+
+impl<T> From<FromUintError<T>> for StorageError {
+    fn from(_value: FromUintError<T>) -> Self {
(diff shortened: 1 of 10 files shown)


src/types/errors.rs
```diff
@@ -0,0 +1,23 @@
+//! Errors stemming from conversions between types used throughout the system.
+
+use stylus_sdk::alloy_primitives::ruint::FromUintError;
+
+/// Errors stemming from working with Solidity-ABI-compatible storage types
+pub enum StorageError {
+    /// An error that occurred while converting between a Solidity-ABI-compatible storage type and a Rust type
+    Conversion,
```
joeykraut: nit: maybe `TypeConversion`, I'm thinking if I see an error log, It'd be more understandable to see `StorageError::TypeConversion` than `StorageError::Conversion`

src/types/solidity_types.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+
+use crate::{
+    constants::{NUM_SELECTORS, NUM_WIRE_TYPES},
+    types::G1Affine,
+};
+
+use super::{errors::StorageError, G2Affine, ScalarField, VerificationKey};
+
+#[inline]
```
joeykraut: I'd say we remove inlines until we know they provide a performance boost. Couple reasons:
1. In rust, just using `#[inline]` is a suggestion, it doesn't guarantee the inline is made. `rustc` is good at choosing when to inline, so the suggestion is unlikely to make any salient difference here.
2. If we're concerned about binary size, inlining generally will increase that.

joeykraut (approved):
Looking good, I like the direction you're taking the abstractions

PR #96, opened 2023-10-11
meta: initialize cargo workspace, write transcript test

akirillo:
This PR modularizes the repo into two library crates: one responsible for all of the core logic of the smart contracts, and the other responsible for actually defining the smart contracts and invoking their functionality in a way that is dependent on the runtime.

This makes unit testing more straightforward as well, since dependencies can be managed separately and we can test the core crate without running into linker issues.

To that end, I've made a test ensuring transcript consistency between the implementation here, and the Jellyfish implementation. In this test, we delegate the `keccak` functionality to the `sha3` crate, instead of `stylus_sdk`. This means that we will have transcript equivalency as long as the `stylus_sdk::keccak` is consistent with the `sha3` implementation, an assumption we are already beholden to as this is the implementation used prover-side.


contracts-core/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "contracts-core"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
joeykraut: nit: remove this

joeykraut (approved):

contracts-core/src/transcript/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+use alloc::vec::Vec;
+use ark_ff::PrimeField;
+use ark_serialize::CanonicalSerialize;
+use core::result::Result;
+
+use crate::types::{Challenges, Proof, ScalarField, VerificationKey};
+
+use self::errors::TranscriptError;
+
+pub trait Transcript {
```
joeykraut: It may make more sense to scope the trait to the hasher instead of the full transcript. That way we mock out less of the actual transcript for testing. I'll let you decide though

akirillo: i like this, will give it a stab

akirillo: ok, my approach was to get rid of the trait altogether, and instead implement directly on a `Transcript` struct which now has a `PhantomData` field generic over `H: TranscriptHasher`. The rationale there was that the only part that one should be able to change is the hash method, not the actual stateful transcript struct

PR #97, opened 2023-10-12
contracts-core: verifier: first-pass vanilla Plonk verifier implementation

akirillo:
This PR introduces a first-pass, untested implementation of the vanilla Plonk verifier algorithm.

The code is unoptimized, with some comments dropped in pointing out future optimizations. Generally, the code was written to match the specification in the paper, and with inspiration from the Jellyfish implementation, particularly for parts of the algorithm that must be generalized to support custom gates (i.e., different numbers of wire types).

The `verify` algorithm is generic over `G: G1EcArithmetic`, which allows differing implementations of EC addition, scalar multiplication, and pairings.

No tests are written yet, as this will require some scaffolding for generating vanilla Plonk circuits and proofs.


contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
+//! A Plonk verifier
```
joeykraut: nit: This should be more descriptive, think about what you'd like to know if you were on the module docs for this page. A good doc comment would explain some of the interfaces and ways in which the code is generic

contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+
+use crate::{
+    constants::NUM_WIRE_TYPES,
+    transcript::{Transcript, TranscriptHasher},
+    types::{G1Affine, G2Affine, Proof, ScalarField, VerificationKey},
+};
+
+use self::errors::VerifierError;
+
+pub trait G1EcArithmetic {
```
joeykraut: What is the object that implements this trait? This either needs more documentation or a different interface. The way the interface is written (i.e. without reference to `self`) it seems like a static object that does EC calculations of some sort. I imagine this is where the precompile fits in? 

This is a bit of an odd abstraction in its current state, I suggest one of two options:
- Redefine this to be more object centric: i.e. implementations also implement `Add<Self> + Mul<Scalar> + Pairing<...>`. Then use different implementations for different backends (e.g. for mocking vs prod)
- Rename this trait to `G1Backend` or `G1ArithmeticBackend` and add doc comments to convey the way in which this fits into the orientation of abstractions

contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+        a_2: &G1Affine,
+        b_2: &G2Affine,
+    ) -> Result<bool, VerifierError>;
+}
+
+pub struct Verifier {
+    /// The verification key for the circuit
+    pub vkey: VerificationKey,
+    /// The evaluation domain for the PIOP
+    pub domain: Radix2EvaluationDomain<ScalarField>,
```
joeykraut: Why do we store this in the verifier struct directly? Will this be in contract storage and generated when the contract is initialized (i.e. reused across proofs)? If so, lets document why. If not, it's cleaner to just generate the domain in the `verify` method

contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+        // Similarly to the assumptions for step 2, the membership of the public inputs in the scalar field
+        // should be enforced by the type system.
+        if public_inputs.len() != self.vkey.l as usize {
+            return Err(VerifierError::InvalidPublicInputs);
+        }
+
+        let mut transcript = Transcript::<H>::new();
+
+        // Step 4: Compute the challenges
+        // TODO: Do we need to pass in an extra transcript init message here?
```
joeykraut: We should just match `jellyfish` wrt this question. Otherwise the prover & verifier transcripts will differ

contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+            self.vkey.selector_comms[1],
+            self.vkey.selector_comms[2],
+            self.vkey.selector_comms[3],
+        ];
+        let line_1_result = line_1_scalars
+            .iter()
+            .zip(line_1_points.iter())
+            .try_fold(self.vkey.selector_comms[4], |acc, (scalar, point)| {
+                G::ec_add(&acc, &G::ec_scalar_mul(scalar, point)?)
+            })?;
```
joeykraut: To simplify this code, you could add `msm` to the trait and give it a default implementation like the one you have here. Reading the current code vs `G::msm` is a big difference in a very complex function.

contracts-core/src/verifier/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+                + (proof.wire_evals.iter().zip(self.vkey.k.iter()).fold(
+                    challenges.alpha,
+                    |acc, (wire_eval, k)| {
+                        acc * (wire_eval
+                            + &(challenges.beta * k * challenges.zeta)
+                            + challenges.gamma)
+                    },
+                ))),
+            &proof.grand_product_comm,
+        )?;
```
joeykraut: I'd prefer if we split some of these steps out into more readable functions. For example you could have:
```rust
fn step_9(...) {
    let mut res = step9_line1(...);
    res += step9_line2(...);
    ...
}

fn step9_line_2(...) {
    let res = lagrange_1_eval * alpha * alpha;
    res += ...;
}
```

Also it would be ideal, but not necessary, if the terms were added in the reading order they are in the paper.

joeykraut (commented):
I want us to take another pass at this; it's important to nail down a super slick verifier implementation as a solid foundation for the rest of the contract. To that end a few general suggestions for readability:
1. Break more logic out into functions and try not to one-line each step. It's okay if the code is verbose: we will let the compiler compress it and we can optimize it down the line if it's a problem anyways.
2. If we're going to tie variable names to their steps in the paper, we should make the _math_ look like it does in the paper. Specifically fewer large chains of iterators and more direct use of std::ops `+`, `-`

As much as we can move towards readability in this part of the codebase, we should.

contracts-core/src/verifier/mod.rs
akirillo: good eye, this is inspired by the Jellyfish implementation, which keeps it in the verifier struct. they do so because they need to reference the `domain` in different public methods, i'll opt for removing it and generating in `verify` as you mentioned

contracts-core/src/verifier/mod.rs
akirillo: I like the first option here so that we can overload operators and make the code more readable, as you mention elsewhere. I was fitting a bit too tightly to "this is where you call the exact same function as a precompile", but this can be fit into a more object-oriented approach as well. I'll take a stab

contracts-core/src/verifier/mod.rs
akirillo: ack, will split each step into a separate function, and once i make use of operator overloading i can do the terms in reading order (as opposed to bumping some stuff to the initial term of `fold`).

it's worth noting though that some parts of this will still need `fold`s and won't read quite like the paper because of preemptive support for custom gates, which may introduce more wire types, causing us to iterate over e.g. `wire_evals` or `permutation_evals` instead of indexing directly.

given this, and my alignment with the notion of having a readable codebase first, and an extensible one later, I actually err on the side of locking down fan-in 2, fan-out 1 gates for now, which would make a lot of this confusing iteration go away and exactly match the paper. we could still support custom gates this way, we would just have to parameterize `NUM_SELECTORS`. thoughts?

contracts-core/src/verifier/mod.rs
joeykraut: I actually prefer the second option, it's simpler

contracts-core/src/verifier/mod.rs
joeykraut: Yeah that seems like a sensible simplification for the time being. I don't think we will use fan-in-k in circuit definitions for a v0

contracts-core/src/verifier/mod.rs
akirillo: Ok, I ended up going with option 2 after all, it was indeed simpler than making all the types & methods generic over `G1`, `G2` and `S` (for scalar field), and with the addition of the `msm` method as you suggested it was sufficiently readable (instead of using `+`, `-` std ops).

That said, it's 1) slightly less efficient, due to the usage of `ScalarField::one` in some MSMs invoking extra EC ops, and 2) we may want to implement these generics down the line as we move away from the specific `G1Affine`, `G2Affine`, `ScalarField` types (which will likely happen)

contracts-core/src/verifier/mod.rs
akirillo: I decided to hardcode both the number of wires (to just $a, b, c$) _and_ the number of selectors (to just $q_L, q_R, q_O, q_M, q_C$) for now to match the paper - we can add custom gates later when we implement them

joeykraut: sounds good

joeykraut (approved):
Much more readable, flows very nicely

PR #99, opened 2023-10-16
contracts-core: verifier: add verification tests & fix implementation bugs

akirillo:
This PR introduces basic tests asserting that a valid proof verifies, and that an invalid one does not.

The PR also includes the fixes necessary to ensure that these tests pass.

### DIFF
--- a/contracts-core/src/verifier/mod.rs
+++ b/contracts-core/src/verifier/mod.rs
@@ -82,7 +82,7 @@ pub fn verify<H: TranscriptHasher, G: G1ArithmeticBackend>(
         .ok_or(VerifierError::Inversion)?
         * zero_poly_eval;
 
-    let lagrange_1_eval = step_6(zero_poly_eval_div_n, &domain, &challenges);
+    let lagrange_1_eval = step_6(zero_poly_eval_div_n, &challenges);
 
     let pi_eval = step_7(
         lagrange_1_eval,
@@ -96,9 +96,9 @@ pub fn verify<H: TranscriptHasher, G: G1ArithmeticBackend>(
 
     let d_1 = step_9::<G>(zero_poly_eval, lagrange_1_eval, vkey, proof, &challenges)?;
 
-    // Increasing powers of v, starting w/ v
-    let mut v_powers = [challenges.v; NUM_WIRE_TYPES * 2 - 1];
-    for i in 1..NUM_WIRE_TYPES * 2 - 1 {
+    // Increasing powers of v, starting w/ 1
+    let mut v_powers = [ScalarField::one(); NUM_WIRE_TYPES * 2];
+    for i in 1..NUM_WIRE_TYPES * 2 {
         v_powers[i] = v_powers[i - 1] * challenges.v;
     }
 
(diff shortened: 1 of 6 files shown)


contracts-core/src/verifier/mod.rs
```diff
@@ -97,8 +97,8 @@ pub fn verify<H: TranscriptHasher, G: G1ArithmeticBackend>(
     let d_1 = step_9::<G>(zero_poly_eval, lagrange_1_eval, vkey, proof, &challenges)?;
 
     // Increasing powers of v, starting w/ v
```
joeykraut: Update this comment reflecting the change to start with 1

contracts-core/src/verifier/mod.rs
```diff
@@ -141,17 +141,12 @@ fn step_5(domain: &Radix2EvaluationDomain<ScalarField>, challenges: &Challenges)
...
-    let Radix2EvaluationDomain {
-        group_gen: omega, ..
-    } = domain;
+fn step_6(zero_poly_eval_div_n: ScalarField, challenges: &Challenges) -> ScalarField {
     let Challenges { zeta, .. } = challenges;
 
-    zero_poly_eval_div_n * *omega / (*zeta - *omega)
+    // For some reason, Jellyfish uses 1, instead of omega, as in the paper,
+    // for the first Lagrange polynomial
+    zero_poly_eval_div_n / (*zeta - ScalarField::one())
```
joeykraut: This is just the case because we've "taken" omega to the group order, or equivalently to 0; completing the evaluation over the cyclic group. This is equivalent to what plonk does I believe. Considering we know why, let's update this comment, removing the obscurity

contracts-core/src/verifier/mod.rs
```diff
@@ -336,7 +335,6 @@ fn step_9_line_4<G: G1ArithmeticBackend>(
     // zeta^{n+2}, as opposed to zeta^n, as in the paper.
     // This is in order to "achieve better balance among degrees of all splitting
     // polynomials (especially the highest-degree/last one)"
-    // (As indicated in the doc comment here: https://github.com/EspressoSystems/jellyfish/blob/main/plonk/src/proof_system/prover.rs#L893)
```
joeykraut: Why'd we remove this? I feel it's helpful documentation?

contracts-core/src/verifier/mod.rs
```diff
@@ -435,3 +433,146 @@ fn step_12<G: G1ArithmeticBackend>(
...
+    fn unwrap_commitments<const N: usize>(comms: &[Commitment<Bn254>]) -> [G1Affine; N] {
+        comms
+            .iter()
+            .map(|c| c.0)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap()
+    }
+
+    fn gen_circuit(n: usize) -> Result<PlonkCircuit<ScalarField>, PlonkError> {
```
joeykraut, line 502: Would be good to expand this circuit just a bit to include multiplication, this better tests the gate identity

joeykraut (approved):
Lgtm

PR #101, opened 2023-10-18
contracts-stylus: verifier: precompile-aware g1 arithmetic backend

akirillo:
This PR introduces an implementation of the `G1ArithmeticBackend` for Stylus contracts, i.e. it calls out to the `ecAdd`, `ecMul`, and `ecPairing` precompiles. This includes de/serialization logic of G1/G2 points & scalars, and some test scaffolding. Namely, some quick unit tests for de/serialization, a test contract wrapping the calls to `G1ArithmeticBackend::ec_add`, `G1ArithmeticBackend::ec_scalar_mul`, `G1ArithmeticBackend::ec_pairing_check`, and a quick script in a new `integration` crate that calls said methods on the test contract. All these test pass.

This PR also includes an implementation of the `verify` method on the `Verifier` contract, but I was unable to test this as the resulting WASM binary is too large. That's an issue for the next PR.

I'm trying to keep the abstractions light and a bit hacky for now, as I don't want to get bogged down before it's clear how the underlying type system will change. If anything looks particularly unwieldy let me know, of course, but it's explicitly not my focus to be writing ergonomic code right now.


contracts-core/src/verifier/mod.rs
```diff
@@ -94,19 +105,26 @@ pub fn verify<H: TranscriptHasher, G: G1ArithmeticBackend>(
 
     let r_0 = step_8(pi_eval, lagrange_1_eval, &challenges, proof);
 
-    let d_1 = step_9::<G>(zero_poly_eval, lagrange_1_eval, vkey, proof, &challenges)?;
+    let d_1 = step_9::<G>(
+        backend,
```
joeykraut: It may read better if we encapsulated this `backend` into an object `Verifier` and use `self.step_9` etc. This has two benefits:
1. The `Verifier` can hold the backend, cleaner than passing it around
2. The `Verifier` can encapsulate the backend generic `G` to avoid coloring all the methods with this generic

contracts-stylus/src/constants.rs
```diff
@@ -0,0 +1,6 @@
+//! Constants used throughout the contracts
+
+pub const EC_ADD_ADDRESS_LAST_BYTE: u8 = 6;
+pub const EC_MUL_ADDRESS_LAST_BYTE: u8 = 7;
+pub const EC_PAIRING_ADDRESS_LAST_BYTE: u8 = 8;
+pub const PAIRING_CHECK_RESULT_LAST_BYTE_INDEX: usize = 31;
```
joeykraut: Let's add doc comments here, it's unclear at first glance what these are used for

contracts-stylus/src/test_contracts.rs
```diff
@@ -0,0 +1,74 @@
...
+};
+use stylus_sdk::prelude::*;
+
+use crate::utils::{ec_add_impl, ec_pairing_check_impl, ec_scalar_mul_impl};
+
+#[solidity_storage]
+#[entrypoint]
+struct TestContract;
+
+impl G1ArithmeticBackend for TestContract {
```
joeykraut: See my comment on the implementation methods below, I don't think implementing the backend trait on the contract itself makes sense from an object orientation perspective. I'd instead wrap the implementation methods in a `EvmPrecompileBackend` (or similar) unit struct. This way it can be used for both test and production contracts as well

contracts-stylus/src/test_contracts.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        Ok(())
+    }
+
+    pub fn test_pairing(&mut self) -> Result<(), Vec<u8>> {
+        let a_1 = G1Affine::generator();
+        let b_1 = G2Affine::generator();
+
+        let a_2 = G1Affine::generator();
+        let b_2 = G2Affine::generator();
```
joeykraut: Can we make these values random, or just not the group generator?

contracts-stylus/src/utils.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+impl From<BackendError> for VerifierError {
+    fn from(_value: BackendError) -> Self {
+        VerifierError::BackendError
+    }
+}
+
+/// Calls the `ecAdd` precompile with the given points, handling de/serialization
+pub fn ec_add_impl<S: TopLevelStorage>(
```
joeykraut: From above, I'd encapsulate these methods in a `EvmPrecompileBackend` struct and implement the `G1ArithmeticBackend` trait on this instead. It makes more sense from an object orientation perspective and can be used in multiple places

joeykraut (approved):
Looks good, couple suggestions around object orientation

contracts-stylus/src/utils.rs
akirillo: Hmm, the thing is we need a mutable reference to the contract struct, which itself implements `TopLevelStorage`. I could do this through a `EvmPrecompileBackend` struct which holds a `&mut dyn TopLevelStorage`, but this seems a lot more unwieldy to me (and with runtime cost), not to mention it would introduce lifetime management to the picture

joeykraut: The runtime cost here is probably negligible: we are running in wasm (a VM) so everything is dynamic lookup (ish) anyways. We can cross the optimization bridge when we get there with a solid foundation.

I think there is a solution to be found that more cleanly encapsulates the shared functionality. Storing a pointer is not that bad, and we can use smart pointer structs if we like. Another option may be passing in a `Context` of sorts that can encapsulate runtime information, etc.

These are suggestions; feel free to take none of them, but I don't think the current abstraction is sufficient.

akirillo: nvm, got a nicer way to make this work

PR #102, opened 2023-10-22
contracts-core: serde: custom serialization avoiding Arkworks into_bigint

akirillo:
This PR builds out the `serde` module (previously called `utils`) to introduce custom de/serialization traits for EVM precompiles & transcript operations. The motivation for this is to avoid Arkworks' `CanonicalSerialize` / `CanonicalDeserialize`, which invoke functionality with tons of forced inlining & loop unrolling, ballooning our binary size.

There are different serde traits for the different contexts (precompiles, transcript, etc.). The rationale for this is included in the module comment.

Additionally, this PR stubs out a `CalldataSerializable` trait which will be implemented in a future PR, this is necessary to parse calldata into `VerificationKey` and `Proof` types to test the verifier as a smart contract (using `CanonicalDeserialize` links in the costly code paths).

All existing tests pass, including transcript equivalency tests & precompile unit tests. Additionally, an optimized release build of the verifier contract is now brotli-compressed down to 14.8 kb.


contracts-core/src/serde.rs
```diff
@@ -0,0 +1,386 @@
...
+
+// ---------------------
+// | TRAIT DEFINITIONS |
+// ---------------------
+
+// --------------
+// | Precompile |
+// --------------
+
+pub trait PrecompileSerializable {
```
joeykraut: Why so many different deserialization traits? Do we expect to be deserializing implementing types differently by target? If not, code could be simpler until that assumption is violated

joeykraut: I see, it looks like below we do this to switch endianness, not sure that warrants a whole new trait.

I think of serialization as mapping data types to byte buffers. How we rearrange that into inputs may be more simply expressed directly rather than through traits.

contracts-core/src/serde.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+}
+
+// -----------
+// | HELPERS |
+// -----------
+
+/// Converts a field element into an Arkworks `BigInt`.
+///
+/// Copied from Arkworks, but omitting loop unrolling and forced inlining.
```
joeykraut: Let's add a code link here for better attribution

joeykraut (approved):
Lgtm, just raising a discussion around abstraction boundaries. Feel free to merge around it if you think the solution is appropriate

contracts-core/src/serde.rs
akirillo: I guess in my case, I see serialization as prepping a type for external consumption, and here the different traits are for the different contexts that the type is consumed in. I had also thought about having a single trait, and using wrapper types to differentiate implementations of the trait for precompiles, transcript, etc.

The rationale was mostly just to keep downstream code ergonomic, e.g. to be able to just call `type.serialize()` without having to wrap / unwrap with another type or call another helper function to ensure the serialization matches the expected format. We can also get some stronger type safety out of this, e.g. compiler warnings to the tune of "this type doesn't implement `PrecompileSerializable`", etc.

Finally, I don't want a single generalized serialization scheme to be constrained by how the Jellyfish transcript or EVM precompiles expect input.

PR #104, opened 2023-10-25
contracts-core: simplify serde using arkworks patch

akirillo:
This PR simplifies the `Serializable` and `Deserializable` trait implementations after patching Arkworks with a [fork](https://github.com/renegade-fi/algebra/tree/bin-opt-feature) that removes all inlining and loop unrolling (if the `bin-opt` feature is enabled)

### DIFF
--- a/contracts-core/src/constants.rs
+++ b/contracts-core/src/constants.rs
@@ -15,3 +15,9 @@ pub const HASH_OUTPUT_SIZE: usize = 32;
 /// The number of bytes to represent field elements of the base or scalar fields for the G1 curve group,
 /// as well as the base field which is extended for the G2 curve group
 pub const FELT_BYTES: usize = 32;
+
+/// The number of u64s it takes to represent a field element
+pub const NUM_U64S_FELT: usize = 4;
+
+/// The number of public inputs in the verifier testing circuit
+pub const NUM_PUBLIC_INPUTS: usize = 0;
--- a/contracts-core/src/serde.rs
+++ b/contracts-core/src/serde.rs
@@ -1,23 +1,23 @@
 //! Custom de/serialization logic used to serialize objects into byte arrays
 //! for use in EVM precompiles, transcript operations, and calldata.
-//!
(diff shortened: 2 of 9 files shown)


contracts-core/src/constants.rs
```diff
@@ -15,3 +15,9 @@ pub const HASH_OUTPUT_SIZE: usize = 32;
 /// The number of bytes to represent field elements of the base or scalar fields for the G1 curve group,
 /// as well as the base field which is extended for the G2 curve group
 pub const FELT_BYTES: usize = 32;
+
+/// The number of u64s it takes to represent a field element
+pub const NUM_U64S_FELT: usize = 4;
+
+/// The number of public inputs in the verifier testing circuit
+pub const NUM_PUBLIC_INPUTS: usize = 0;
```
joeykraut, line 23: We should relocate this to testing code if it's only used for testing, it doesn't make much sense to have it here IMO

contracts-core/src/serde.rs
```diff
@@ -1,105 +1,108 @@
...
+#[derive(Debug)]
+pub struct SerdeError;
+
 pub trait Serializable {
     /// Serializes a type into a vector of bytes
     fn serialize(&self) -> Vec<u8>;
 }
 
 pub trait Deserializable {
+    const SER_LEN: usize;
```
joeykraut, line 28: Why do we have this as an associated constant? This prevents our ability to implement this trait on variable sized structs (vectors) for example. A length prefix might make more sense in this case

joeykraut (approved):

contracts-core/src/serde.rs
akirillo: an implementation that accounts for dynamically-sized types introduces a fair bit of complexity that I feel would only slow us down right now, given it's not immediately necessary. I also think we can get away with having all lengths be parameterized by constants

contracts-core/src/constants.rs
akirillo: ah, this gets used in the next PR

PR #105, opened 2023-10-25
integration: verifier test & testing CLI

akirillo:
This PR introduces a lightweight CLI for running integration tests, taking some inspiration from the way the Starknet client integration tests are run in the relayer repo.

Additionally, it implements a basic integration test for the verifier, asserting that a valid proofs verifies and an invalid one does not.

All of the test, including existing unit and integration tests, pass.

Much of the size of this diff is lockfile and me reorganizing existing code so that test helpers can be shared.


integration/src/cli.rs
```diff
@@ -0,0 +1,32 @@
+//! Definition of the CLI arguments for integration tests
+
+use crate::constants::{DEFAULT_DEVNET_HOSTPORT, DEFAULT_DEVNET_PKEY};
+use clap::{Parser, ValueEnum};
+
+/// CLI tool for running integration tests against a running devnet node.
+///
+/// Assumes that the contracts invoked in the tests have already been deployed to the devnet.
+#[derive(Parser)]
+pub(crate) struct Cli {
```
joeykraut, line 10: Is the idea that we run a devnet sequencer with the contracts pre-deployed?

joeykraut (approved):

integration/src/cli.rs
akirillo: yes, that's the assumption

PR #106, opened 2023-10-26
contracts-stylus: initialize darkpool contract w/ nullifier set

akirillo:
This PR initializes the darkpool contract, initially as a thin wrapper around nullifier set functionality. This functionality is implemented directly in `contracts-stylus` using the Stylus SDK, as opposed to having core logic be in `contracts-core`, since it is effectively direct calls on the Stylus SDK for interacting with a storage mapping, and any abstraction on top of this would be distracting and unnecessary.

Additionally, a simple integration test for nullifier set functionality was added.


contracts-stylus/src/darkpool.rs
```diff
@@ -0,0 +1,30 @@
...
+struct DarkpoolContract {
+    /// The set of wallet nullifiers, representing a mapping from a nullifier
+    /// (which is a Bn254 scalar field element serialized into 32 bytes) to a
+    /// boolean indicating whether or not the nullifier is spent
+    nullifier_set: StorageMap<B256, StorageBool>,
+}
+
+#[external]
+impl DarkpoolContract {
+    pub fn is_nullifier_spent(&self, nullifier: B256) -> Result<bool, Vec<u8>> {
```
joeykraut, line 23: Although they're simple, we should have doc comments for at least all external methods

joeykraut (approved):

PR #110, opened 2023-10-31
contracts-core: serde: add `ScalarSerializable` trait & impls for structured statements

akirillo:
This PR is the first step towards having structured statements for our proofs. This is important as we need to access statement elements in pre- and post-verification functionality, such as verifying ECDSA signatures, or inserting commitments into the Merkle tree. This PR copies over the statement types from the old Cairo repo, and adds a new trait `ScalarSerializable`.

The intended usage is that when calling a wallet interaction method such as `new_wallet`, the serialization of the statement into calldata goes in two stages:
1. Serialize the statement into an array of scalars using `ScalarSerializable`
2. Serialize the array of scalars into an array of bytes using the existing `Serializable` trait

This way, we can deserialize the structured statements from calldata (which is a byte array), but also directly pass the calldata to the verifier (which expects the public inputs to be an array of scalars, serialized into bytes).

I save the implementation of `ScalarDeserializable` and `Deserializable` on the statement types, as well as testing, for a future PR, as this one is growing quite large (albeit quite repetitive)

### DIFF
--- a/common/src/constants.rs
+++ b/common/src/constants.rs
@@ -19,6 +19,9 @@ pub const FELT_BYTES: usize = 32;
 /// The number of u64s it takes to represent a field element
 pub const NUM_U64S_FELT: usize = 4;
 
+/// The number of bytes it takes to represent an unsigned 256-bit integer
+pub const NUM_BYTES_U256: usize = 32;
+
 /// The number of public inputs in the verifier testing circuit
 pub const NUM_PUBLIC_INPUTS: usize = 0;
 
--- a/common/src/constants.rs
+++ b/common/src/constants.rs
@@ -21,3 +21,6 @@ pub const NUM_U64S_FELT: usize = 4;
 
 /// The number of public inputs in the verifier testing circuit
 pub const NUM_PUBLIC_INPUTS: usize = 0;
+
+/// The number of secret-shared scalars it takes to represent a wallet
+pub const WALLET_SHARES_LEN: usize = 0;
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -100,7 +100,7 @@ pub struct PublicSigningKey {
 pub struct ValidWalletCreateStatement {
     /// The commitment to the private secret shares of the wallet
     pub private_shares_commitment: ScalarField,
-    /// The public secret shares of the wallet
+    /// The blinded public secret shares of the wallet
     pub public_wallet_shares: [ScalarField; WALLET_SHARES_LEN],
 }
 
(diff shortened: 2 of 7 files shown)


common/src/constants.rs
```diff
@@ -19,5 +19,11 @@ pub const FELT_BYTES: usize = 32;
...
 pub const NUM_U64S_FELT: usize = 4;
 
+/// The number of bytes it takes to represent an unsigned 256-bit integer
+pub const NUM_BYTES_U256: usize = 32;
+
 /// The number of public inputs in the verifier testing circuit
 pub const NUM_PUBLIC_INPUTS: usize = 0;
+
+/// The number of secret-shared scalars it takes to represent a wallet
+pub const WALLET_SHARES_LEN: usize = 0;
```
joeykraut, line 29: Do we really want to set this as a constant, that implies we redeploy contracts every time we add fields to a wallet, even if they imply no circuit constraints

common/src/types.rs
```diff
@@ -77,3 +75,100 @@ pub struct Challenges {
...
+    pub public_wallet_shares: [ScalarField; WALLET_SHARES_LEN],
+}
+
+/// Statement for `VALID_WALLET_UPDATE` circuit
+pub struct ValidWalletUpdateStatement {
+    /// The nullifier of the old wallet's secret shares
+    pub old_shares_nullifier: ScalarField,
+    /// A commitment to the new wallet's private secret shares
+    pub new_private_shares_commitment: ScalarField,
+    /// The public secret shares of the new wallet
```
joeykraut: Let's mention that these are blinded, to make absolutely clear

common/src/types.rs
```diff
@@ -77,3 +75,100 @@ pub struct Challenges {
...
+    pub party1_send_balance_index: u64,
+    /// The index of the balance received by the second party in the settlement
+    pub party1_receive_balance_index: u64,
+    /// The index of the second party's matched order
+    pub party1_order_index: u64,
+}
+
+/// Represents the outputs produced by one of the parties in a match
+pub struct MatchPayload {
+    pub wallet_blinder_share: ScalarField,
```
joeykraut, line 170: nit: no doc comments on these?

contracts-core/src/serde.rs
```diff
@@ -246,6 +251,203 @@ impl Deserializable for Proof {
...
+    fn to_scalars(&self) -> Vec<ScalarField> {
+        vec![ScalarField::from(*self)]
+    }
+}
+
+impl ScalarSerializable for u64 {
+    fn to_scalars(&self) -> Vec<ScalarField> {
+        vec![ScalarField::from(*self)]
+    }
+}
```
joeykraut: Why not:
```rust
impl<T: Into<ScalarField>> ScalarSerializable for T {
    fn to_scalars(&self) -> Vec<ScalarField> {
        vec![self.into()]
    }
}
```

joeykraut (approved):

common/src/constants.rs
akirillo: hmm, yeah true... might be about time that I tackle variable-length type serialization

akirillo: ok, think I have a design for this, but will implement it later in one fell swoop along side dynamic-sized circuit & statement support in the verifier

PR #116, opened 2023-11-02
contracts-core, contracts-stylus: verifier precompile w/ var-length public inputs

akirillo:
This PR changes the verifier contract to be a "minimal stylus entrypoint", in that it effectively acts as a precompile. The contract gets static-called directly without exposing an ABI-encoded interface. This brings the binary size down by avoiding ABI encoding/decoding code paths.

This became necessary when introducing variable-length inputs, as the code bloat from deserializing a vector of scalars just pushed the verifier contract over the limit.

As such, this PR also includes the changes necessary for variable-length public inputs, and the updates to the Darkpool contract to ingest these changes.

In order to test the verifier, I've introduced a test contract which exposes an interface, this makes it much more straightforward to interact with in testing code. All tests pass, save for the Darkpool verification test which has been removed in anticipation of tests with circuits that match the statements expected by the Darkpool.

### DIFF
--- a/contracts-stylus/src/verifier.rs
+++ b/contracts-stylus/src/verifier.rs
@@ -1,15 +1,13 @@
 //! The verifier smart contract, responsible for verifying Plonk proofs.
 
-use alloc::vec::Vec;
+use alloc::{vec, vec::Vec};
 use common::constants::HASH_OUTPUT_SIZE;
-use common::{
-    constants::NUM_PUBLIC_INPUTS,
-    types::{Proof, ScalarField, VerificationKey},
-};
+use common::types::VerificationBundle;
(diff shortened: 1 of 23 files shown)


contracts-stylus/src/verifier.rs
```diff
@@ -20,30 +18,19 @@ impl TranscriptHasher for StylusHasher {
     }
 }
 
-#[solidity_storage]
+/// Verify the given proof, using the given public inputs and the stored verification key
```
joeykraut: nit: the verification key is not stored unless I'm misreading

joeykraut (approved):
lgtm

PR #117, opened 2023-11-03
contracts-stylus: darkpool: accept vkey as calldata and check against stored hash

akirillo:
This PR changes the implementation of the Darkpool to no longer store verification keys in storage, but instead store their hashes. For any methods that require verification, a verification key is supplied as calldata, and is checked against the stored hash.

While this change also saves gas (hashing the verification key is cheaper than reading it in from (cold) storage), the immediate need is actually because writing a variable-length byte array to storage indirectly invokes the currently-most-bloated code path in the `stylus_sdk`: checking `stylus_sdk::StorageBytes::len` calls `ruint:div` which calls out to some forced-inline functions.

I prefer this approach to patching in a fork of `ruint` that we have to maintain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,13 +50,14 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


akirillo:
closing after internal discussion regarding costs associated w this approach (calldata is priced in L1 gas, so this is not cheaper), and in favor of #118 as a means of taming binary size

PR #118, opened 2023-11-03
common: custom_serde: scalar serde directly to/from bigint

akirillo:
This PR changes the way that we deserialize field elements from bytes, and the way that we serialize Ethereum addresses / `uint256`s to scalars. Concretely, we sidestep much of the unnecessary code bloat linked in by calling `from_be_bytes_mod_order` in Arkworks, and instead form scalars directly from the underlying `BigInt` type.

The reason for this is to decrease binary size. This change decreases the Brotli-compressed size of the darkpool contract from 24.4kb -> 20.8kb.

All existing tests pass. Deserialization of a field element from bytes is covered by the existing tests, and testing serialization of the Solidity types to scalars is deferred for now. Unit testing this would imply creating a trait to deserialize a type from scalars, which is not functionality that we currently (or, I think, ever) need.

### DIFF
--- a/common/src/custom_serde.rs
+++ b/common/src/custom_serde.rs
@@ -7,10 +7,9 @@ use alloy_primitives::{Address, U256};
 use ark_ec::{short_weierstrass::SWFlags, AffineRepr};
 use ark_ff::{BigInt, BigInteger, MontConfig, PrimeField, Zero};
 use ark_serialize::Flags;
-use core::iter;
 
 use crate::{
-    constants::{FELT_BYTES, NUM_BYTES_U256, NUM_BYTES_U64, NUM_U64S_FELT},
+    constants::{FELT_BYTES, NUM_BYTES_U64, NUM_U64S_FELT},
     types::{
         ExternalTransfer, G1Affine, G1BaseField, G2Affine, G2BaseField, MontFp256,
         PublicSigningKey, ScalarField, ValidCommitmentsStatement, ValidMatchSettleStatement,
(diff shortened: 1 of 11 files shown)


common/src/custom_serde.rs
```diff
@@ -185,140 +194,149 @@ impl BytesDeserializable for G2Affine {
...
+        let low_bigint = bigint_from_le_bytes(
+            &low_bytes
+                .try_into()
+                .map_err(|_| SerdeError::InvalidLength)?,
+        )?;
+        let high_bigint = bigint_from_le_bytes(
+            &high_bytes
+                .try_into()
+                .map_err(|_| SerdeError::InvalidLength)?,
+        )?;
```
joeykraut: These two expressions may be simplified as:
```rust
const HALF: usize = NUM_BYTES_U256 / 2;
let padded: Vec<u8> = bytes.into_iter().chain(iter::repeat(0)).take(NUM_BYTES_U256).collect();
let low_bytes = bigint_from_le_bytes(&bytes[..HALF].try_into().unwrap());
let high_bytes = bigint_from_le_bytes(&bytes[HALF..].try_into().unwrap());
```
probably not verbatim, but you get the gist of what I'm suggesting

common/src/custom_serde.rs
```diff
@@ -185,140 +194,149 @@ impl BytesDeserializable for G2Affine {
...
+
+        Ok(vec![
+            ScalarField::from_bigint(high_bigint).ok_or(SerdeError::ScalarConversion)?,
+            ScalarField::from_bigint(low_bigint).ok_or(SerdeError::ScalarConversion)?,
+        ])
```
joeykraut, line 239: This feels like a bit of a roundabout way to get from U256, here we:
1. Decompose to bytes and pad
2. Restructure those bytes into limbs
3. Reconstruct a BigInt from limbs
4. Turn this into a ScalarField

Why go to bytes, why not just grab the limbs off the `U256` directly. Both `ruint` and `arkworks` use the same radix, so the limbs are exactly the same. This is simpler and less compute

joeykraut (approved):
Suggestions for simplification

common/src/custom_serde.rs
akirillo: I'm mad dumb good call

PR #119, opened 2023-11-03
integration: initialize `update_wallet` test w/ nullifier check

akirillo:
This PR introduces some scaffolding to test the various wallet interaction methods on the darkpool, and leverages this to make a barebones `update_wallet` test that for now, assures that the proof verifies and that the correct nullifier was spent.

### DIFF
--- a/integration/src/tests.rs
+++ b/integration/src/tests.rs
@@ -3,7 +3,6 @@
 use ark_ff::One;
 use ark_std::UniformRand;
 use common::{
-    constants::NUM_BYTES_FELT,
     custom_serde::ScalarSerializable,
     serde_def_types::SerdeScalarField,
     types::{ScalarField, VerificationBundle},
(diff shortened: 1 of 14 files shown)


integration/src/tests.rs
```diff
@@ -83,3 +90,65 @@ pub(crate) async fn test_nullifier_set(
...
+    contract
+        .update_wallet(
+            wallet_blinder_share_bytes,
+            proof_bytes,
+            valid_wallet_update_statement_bytes,
+            public_inputs_signature_bytes,
+        )
+        .send()
+        .await?
+        .await?;
```
joeykraut, line 142: Guessing that you're planning to make this setup logic reusable as we introduce more tests?

joeykraut (approved):
lgtm

integration/src/tests.rs
akirillo: yeah, more scaffolding in the next PR, want to let it take shape first

PR #120, opened 2023-11-04
integration: set up `process_match_settle` test, check nullifiers

akirillo:
This PR introduces additional helpers & scaffolding for the integration tests, and adds an integration test for the `process_match_settle` in which we assert that the correct nullifiers were marked as spent.


joeykraut (approved):
Looks good

PR #121, opened 2023-11-05
contracts-core: crypto: add `EcRecoverBackend` for ECDSA verification

akirillo:
This PR introduces the `EcRecoverBackend`, which smart contracts can use to verify ECDSA signatures. It also implements conversion functionality from a scalar-serialized `PublicSigningKey` (as we would have in the `ValidWalletUpdateStatement`) and an Ethereum address.

The PR includes a unit test of ECDSA verification using native hashing & recovery methods, as well some useful scaffolding for generating a random (valid) secp256k1 keypair and serializing the public key to scalars in the appropriate manner.

Finally, there is some diff bloat due to reorganization of the `test-helpers` crate.

Integration tests leveraging the EVM `ecRecover` precompile are left for the next PR to make review quicker here.


contracts-core/src/verifier/mod.rs
```diff
@@ -55,29 +46,28 @@ pub trait G1ArithmeticBackend {
...
-pub struct Verifier<G: G1ArithmeticBackend, H: TranscriptHasher> {
+pub struct Verifier<G: G1ArithmeticBackend, H: HashBackend> {
     /// The verification key for a specific circuit
     vkey: VerificationKey,
     /// A transcript of the verifier's interaction with the prover,
     /// used for the Fiat-Shamir transform
     transcript: Transcript<H>,
-    /// The backend for elliptic curve arithmetic
-    backend: G,
+    _arithmetic_backend: PhantomData<G>,
```
joeykraut: nit: rust best practice is to call this `_phantom` to avoid confusion as to what it does.

test-helpers/src/crypto.rs
```diff
@@ -0,0 +1,63 @@
...
+use rand::{CryptoRng, RngCore};
+
+pub fn random_keypair<R: CryptoRng + RngCore>(rng: &mut R) -> (SigningKey, PublicSigningKey) {
+    let signing_key = SigningKey::random(rng);
+    let verifying_key = signing_key.verifying_key();
+    let encoded_pubkey_bytes = verifying_key
+        .to_encoded_point(false /* compress */)
+        .to_bytes();
+
+    let num_bytes_u128 = NUM_BYTES_U256 / 2;
```
joeykraut: nit: this should be a constant

joeykraut (approved):

PR #122, opened 2023-11-06
contracts-stylus: `PrecompileEcRecoverBackend` impl + test

akirillo:
This PR introduces a backend for the `ecRecover` precompile, and adds a test to assert that it functions properly.

There is also some reorganization: packaging all the "backends" together in the `common` module, and moving testing logic for the precompile test contract outside of the contract itself (as well as splitting into separate tests)


common/src/backends.rs
```diff
@@ -0,0 +1,66 @@
+//! "Backends" representing functionality that is either delegated to
+//! EVM precompiles, or to native Rust code in tests.
```
joeykraut, line 2: It's be nice to note that the primary use of abstracting here is for mocks

joeykraut (approved):

PR #126, opened 2023-11-07
contracts-stylus: components, darkpool: add ownership checks

akirillo:
This PR adds ownership checks to the relevant methods on the darkpool contract by mirroring the functionality of OpenZeppelin's [`Ownable`](https://docs.openzeppelin.com/contracts/5.x/api/access#Ownable) contract in Stylus.

Additionally, this PR adds tests asserting that each of the protected methods can only be called by the owner.

### DIFF
--- a/contracts-stylus/src/contracts/components/ownable.rs
+++ b/contracts-stylus/src/contracts/components/ownable.rs
@@ -1,12 +1,20 @@
 //! Mirrors OpenZeppelin's `Ownable` contract for access controls:
 //! https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.0/contracts/access/Ownable.sol
+//!
+//! Since Stylus does not yet support constructors, we add extra initialization logic.
 
 use alloc::vec::Vec;
-use stylus_sdk::{alloy_primitives::Address, msg, prelude::*, storage::StorageAddress};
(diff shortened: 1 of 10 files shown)


contracts-stylus/src/contracts/components/ownable.rs
```diff
@@ -0,0 +1,57 @@
...
+impl Ownable {
+    pub fn owner(&self) -> Result<Address, Vec<u8>> {
+        Ok(self.owner.get())
+    }
+
+    pub fn renounce_ownership(&mut self) -> Result<(), Vec<u8>> {
+        self._check_owner()?;
+        self._transfer_ownership(Address::ZERO);
+        Ok(())
+    }
```
joeykraut, line 30: More for my own learning, in what situation would we use a method like this, it's a terminal state right? I'd guess this is for things like burning upgrade keys when we want to make a version of our contract permanent?

joeykraut (approved):

contracts-stylus/src/contracts/components/ownable.rs
akirillo: That would be my assumption as well, yeah. Candidly though, I just mirrored this because it's in the original implementation, deferring changes to the interface for later

PR #127, opened 2023-11-08
contracts-core: merkle: implement Poseidon2 Merkle Tree

akirillo:
This PR introduces an implementation of a Merkle tree, leveraging the Poseidon2 implementation in the relayer.

Because the Arkworks implementation of a Merkle tree is dense (stores all nodes), we implement one here from the ground up, mirroring the old [Cairo implementation](https://github.com/renegade-fi/renegade-contracts/blob/e8a057ed2182d1738e0f38bccb668bd0592c525d/src/merkle.cairo).

The implementation is tested against an Arkworks implementation (and passes).

### DIFF
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -248,16 +248,18 @@ pub struct VerificationBundle {
 /// for the next leaf to be inserted.
 #[serde_as]
 #[derive(Serialize, Deserialize)]
-pub struct SparseMerkleTree<const HEIGHT: usize> {
+pub struct SparseMerkleTree {
+    /// The height of the tree, inclusive of the root
+    pub height: usize,
     /// The next index at which to insert a leaf
     pub next_index: u128,
     /// The current root of the tree
     #[serde_as(as = "ScalarFieldDef")]
     pub root: ScalarField,
     /// The current path of siblings for the next leaf to be inserted
-    #[serde_as(as = "[ScalarFieldDef; HEIGHT]")]
-    pub sibling_path: [ScalarField; HEIGHT],
+    #[serde_as(as = "Vec<ScalarFieldDef>")]
+    pub sibling_path: Vec<ScalarField>,
     /// The path of values in an empty tree
-    #[serde_as(as = "[ScalarFieldDef; HEIGHT]")]
-    pub zeros: [ScalarField; HEIGHT],
+    #[serde_as(as = "Vec<ScalarFieldDef>")]
+    pub zeros: Vec<ScalarField>,
 }
--- a/common/src/types.rs
+++ b/common/src/types.rs
@@ -243,3 +243,21 @@ pub struct VerificationBundle {
     #[serde_as(as = "Vec<ScalarFieldDef>")]
     pub public_inputs: Vec<ScalarField>,
 }
+
+/// A low-memory, append-only Merkle tree that only stores the current path of siblings
+/// for the next leaf to be inserted.
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+pub struct SparseMerkleTree<const HEIGHT: usize> {
+    /// The next index at which to insert a leaf
+    pub next_index: u128,
+    /// The current root of the tree
+    #[serde_as(as = "ScalarFieldDef")]
+    pub root: ScalarField,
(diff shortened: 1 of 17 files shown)


common/src/types.rs
```diff
@@ -243,3 +243,23 @@ pub struct VerificationBundle {
...
     pub public_inputs: Vec<ScalarField>,
 }
+
+/// A low-memory, append-only Merkle tree that only stores the current path of siblings
+/// for the next leaf to be inserted.
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+pub struct SparseMerkleTree {
+    /// The height of the tree, inclusive of the root
+    pub height: usize,
```
joeykraut: As an optimization, we can constantize this to avoid reading from storage? The height should not change except with a redeployment anyways. Feel free to leave this out for now, more of a suggestion.

Same applies for the `zeros` field

contracts-core/src/crypto/merkle.rs
```diff
@@ -0,0 +1,179 @@
...
+
+/// Represents a node in the Merkle tree,
+/// including the height and index "coordinates" along with the value
+pub struct Node {
+    pub height: usize,
+    pub index: u128,
+    pub value: ScalarField,
+}
+
+pub trait MerkleTree {
```
joeykraut: Do we need a trait for this? Are we going to realistically have multiple implementations? If not I'd propose removing it for simplicity's sake

contracts-core/src/crypto/merkle.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Create a new Merkle tree of the given height (inclusive of the root),
+    /// w/ zeros as the leaves
+    fn new(height: usize) -> Self;
+    /// Insert a value into the Merkle tree,
+    /// returning the updated opening for the insertion
```
joeykraut: We don't want to return the opening -- if this is for events -- we want to return the values of the internal nodes that have changed.

contracts-core/src/crypto/merkle.rs
```diff
@@ -0,0 +1,179 @@
...
+    // Hash the current leaf with itself and recurse
+    let mut sponge = Poseidon2Sponge::new();
+    let next_leaf = sponge.hash(&[current_leaf, current_leaf]);
+
+    setup_empty_tree(tree, height - 1, next_leaf)
+}
+
+/// Recursive helper for inserting a value into the Merkle tree,
+/// updating the sibling pathway along the way, and returning
+/// the opening for the insertion
```
joeykraut: same here re: above, let's update this comment to not be about the opening but about the values that changed

contracts-core/src/crypto/merkle.rs
```diff
@@ -0,0 +1,179 @@
...
+        index: insert_index,
+        value,
+    }];
+    node_changes.extend(insert_helper(
+        tree,
+        next_value,
+        height - 1,
+        next_index,
+        new_subtree_filled,
+    ));
```
joeykraut: nit: small optimization here, but probably faster to instead push the current change to the recursive result instead of extending a singleton vector. Feel free to ignore, but since our execution environment is so constrained it could be useful. Also doesn't matter the order in which these appear if they're only used for events

joeykraut (approved):
Few small suggestions and re-wordings to make opening vs update distinction

common/src/types.rs
akirillo: I actually started with it being a generic constant, it messes with the constructor method `MerkleTree::new()` when we try to set up `siblings_path` and `zeros` to be arrays of length `HEIGHT - 1` (can't use expressions).

There's certainly easy workarounds but I wanted to move quick, I may give this another go

contracts-core/src/crypto/merkle.rs
akirillo: Re: our discussion, yeah, it was just for code separation, I'll move the impl block into the `common` crate

contracts-core/src/crypto/merkle.rs
akirillo: oh oops yes

PR #128, opened 2023-11-09
contracts-stylus: merkle: add Merkle contract

akirillo:
This PR creates a parameterizable Merkle contract, following the parameterization patterns used [here](https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/examples/erc20/src/erc20.rs) (namely, using a trait w/ associated consts + contract inheritance), and integrates it into the darkpool.

A new integration test has been added for the Merkle contract itself, which passes. Test coverage asserting correct Merkle behavior in the darkpool has not yet been added since the darkpool is once again beyond the binary size limit, which I'll handle in another PR.

Additionally, the Merkle contract should be `Ownable` and `Initializable`, I've left this for the next PR as well, for brevity.

### DIFF
--- a/contracts-stylus/src/contracts/merkle.rs
+++ b/contracts-stylus/src/contracts/merkle.rs
@@ -1,0 +1,62 @@
+//! A Merkle tree smart contract, used to accumulate all of the wallet commitments
+//! in the dark pool.
+
+use alloc::vec::Vec;
+use common::serde_def_types::SerdeScalarField;
+use contracts_core::crypto::merkle::SparseMerkleTree;
+use stylus_sdk::{
+    abi::Bytes,
+    prelude::*,
+    storage::{StorageBool, StorageBytes, StorageMap},
+};
+
+use crate::utils::constants::MERKLE_HEIGHT;
+
+#[solidity_storage]
+#[entrypoint]
+pub struct MerkleContract {
+    /// The serialized merkle tree
+    pub merkle_tree: StorageBytes,
+    /// The current root of the Merkle tree, cached in storage for efficiency
+    pub current_root: StorageBytes,
+    /// The set of historic roots of the Merkle tree
+    pub root_history: StorageMap<Vec<u8>, StorageBool>,
+}
+
+#[external]
+impl MerkleContract {
+    /// Initialize this contract with a blank Merkle tree
+    // TODO: Make an `Initializable` component
+    pub fn init(&mut self) -> Result<(), Vec<u8>> {
+        let merkle_tree = SparseMerkleTree::<MERKLE_HEIGHT>::default();
+        let merkle_tree_bytes = postcard::to_allocvec(&merkle_tree).unwrap();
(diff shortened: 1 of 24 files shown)


contracts-stylus/src/contracts/merkle.rs
```diff
@@ -0,0 +1,104 @@
...
+{
+    /// Initialize this contract with a blank Merkle tree
+    pub fn init(&mut self) -> Result<(), Vec<u8>> {
+        let merkle_tree = SparseMerkleTree::<{ P::HEIGHT }>::default();
+        let merkle_tree_bytes = postcard::to_allocvec(&merkle_tree).unwrap();
+        self.merkle_tree.set_bytes(merkle_tree_bytes);
+        Ok(())
+    }
+    /// Returns the current root of the merkle tree
+    // TODO: Cache this directly in contract storage?
```
joeykraut: What does this comment mean? Isn't it already in contract storage?

contracts-stylus/src/contracts/merkle.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub fn root(&self) -> Result<Bytes, Vec<u8>> {
+        Ok(self.current_root.get_bytes().into())
+    }
+
+    /// Returns whether or not the given root is in the root history
+    pub fn root_in_history(&self, root: Bytes) -> Result<bool, Vec<u8>> {
+        Ok(self.root_history.get(root.0))
+    }
+
+    pub fn insert(&mut self, value: Bytes) -> Result<(), Vec<u8>> {
```
joeykraut, line 57: We should at least have doc comments on all external methods

contracts-stylus/src/contracts/test_contracts/merkle_test_contract.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use alloc::vec::Vec;
+use common::constants::TEST_MERKLE_HEIGHT;
+use stylus_sdk::{abi::Bytes, prelude::*};
+
+use crate::contracts::merkle::{MerkleContract, MerkleParams};
+
+struct TestMerkleParams;
+impl MerkleParams for TestMerkleParams {
+    const HEIGHT: usize = TEST_MERKLE_HEIGHT;
```
joeykraut, line 11: Why not define `TEST_MERKLE_HEIGHT` in this file, it makes more sense here, as it's not used in the production contract

joeykraut (approved):

contracts-stylus/src/contracts/test_contracts/merkle_test_contract.rs
akirillo: I decided to define it in `common` since it's used in a number of places: here, in the unit test against arkworks in `contracts-core`, and in the integration test

contracts-stylus/src/contracts/merkle.rs
akirillo: yup, good catch - ended up implementing this after writing the comment

PR #129, opened 2023-11-09
contracts-stylus: make Poseidon precompile for darkpool

akirillo:
This PR makes some changes to ensure that the Darkpool contract can fit into the required binary size, primarily by breaking out Poseidon hash computation into a separate "custom precompile" contract, aka a minimal Stylus entrypoint, similar to the verifier.

This incurs some small gas overhead (700 gas for a static call, but the serde gas costs will overwhelm this) that I have not measured yet, but will defer the optimization for later.

I don't make use of this precompile in the Merkle contract since that comfortably fits within the binary size limit, anyway.

This also introduces some small custom serialization changes for bottlenecks highlighted by the profile, the tests for which pass.


akirillo:
Closed in favor of the approach in #130

akirillo:
Closed in favor of the approach in #130

PR #130, opened 2023-11-09
contracts-stylus: merkle, darkpool: Merkle contract computes commitment to wallet shares

akirillo:
This PR changes the Merkle contract interface to be more application-specific, in that it accepts a set of wallet shares, and computes their commitment on its own before inserting them into the Merkle tree. This is the only way we use the Merkle tree, and also mitigates binary size bloat in the darkpool contract by removing an invocation of the Poseidon hash function there.

The existing Merkle contract test has been updated to reflect this change, but tests asserting that the Merkle root updates as expected in the invocation of the darkpool method are left for the next PR.

### DIFF
--- a/contracts-core/src/crypto/ecdsa.rs
+++ b/contracts-core/src/crypto/ecdsa.rs
@@ -44,8 +44,14 @@ pub fn pubkey_to_address<H: HashBackend>(pubkey: &PublicSigningKey) -> [u8; NUM_
         .unwrap()
 }
 
+/// Returns the lower 128 bits of a scalar as a big-endian byte array
 fn scalar_lower_128_bytes_be(scalar: &ScalarField) -> Vec<u8> {
     let bigint = scalar.into_bigint();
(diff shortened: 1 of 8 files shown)


contracts-core/src/crypto/ecdsa.rs
```diff
@@ -45,12 +45,10 @@ pub fn pubkey_to_address<H: HashBackend>(pubkey: &PublicSigningKey) -> [u8; NUM_
 }
 
 fn scalar_lower_128_bytes_be(scalar: &ScalarField) -> Vec<u8> {
-    scalar
-        .into_bigint()
-        .to_bytes_be()
-        .into_iter()
-        .skip(NUM_BYTES_FELT / 2)
-        .collect()
+    let bigint = scalar.into_bigint();
```
joeykraut, line 49: Can we add a doc comment explaining what's going on here?

joeykraut (approved):
Nice I like it

PR #131, opened 2023-11-10
contracts-stylus, integration: Correct Merkle operations in darkpool

akirillo:
This PR introduces the necessary changes to the Merkle & darkpool contracts, as well as to the integration tests, to assure that the darkpool is correctly updating the Merkle tree.

Better test coverage, i.e. Merkle root history being correctly updated, etc, is left for a later bolstering of the test suite.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -240,8 +240,25 @@ impl DarkpoolContract {
         let valid_match_settle_statement: ValidMatchSettleStatement =
             postcard::from_bytes(valid_match_settle_statement_bytes.as_slice()).unwrap();
 
-        // TODO: Assert that the Merkle roots for which inclusion is proven in `VALID_REBLIND`
-        // are valid historical roots
+        assert!(self
+            .root_in_history(
+                postcard::to_allocvec(&SerdeScalarField(
+                    party_0_match_payload.valid_reblind_statement.merkle_root
+                ))
+                .unwrap()
+                .into()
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -179,8 +180,13 @@ impl DarkpoolContract {
...
 
-        // TODO: Assert that the Merkle root for which inclusion is proven in `VALID_WALLET_UPDATE`
-        // is a valid historical root
+        assert!(self
+            .root_in_history(
+                postcard::to_allocvec(&SerdeScalarField(valid_wallet_update_statement.merkle_root))
+                    .unwrap()
+                    .into()
+            )
+            .unwrap());
```
joeykraut: Let's make this a helper since it shows up a few times

contracts-stylus/src/contracts/merkle.rs
```diff
@@ -70,15 +73,26 @@ where
...
         Ok(())
     }
 }
 
+impl<P> MerkleContract<P>
+where
+    P: MerkleParams,
+    [(); P::HEIGHT - 1]:,
+{
+    pub fn store_root(&mut self, merkle_tree: &SparseMerkleTree<{ P::HEIGHT }>) {
```
joeykraut: It'd be better to pass in the root directly, it's awkward to pass the whole tree object into this method and adds the need for const generic expressions, which are avoidable.

joeykraut (approved):

PR #132, opened 2023-11-10
contracts-stylus: darkpool, merkle: make darkpool `Initializable` & `delegatecall` Merkle contract

akirillo:
This PR ended up being more of a mixed bag than I'd like, it led to many detours...

Primarly, however, this extends appropriate access controls to the Merkle contract, making it owned by the darkpool.

Technically, we could just `delegatecall` to the Merkle contract and not worry about access controls on the deployed contract itself, but this is a bad idea in the case that we ever point the darkpool to a Merkle implementation that can `selfdestruct`. This is the basis of the pattern that upgradeable contracts (like, in our case, the darkpool) should never `delegatecall` ([reference](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#potentially-unsafe-operations)) - and I err on the side of following best practices, even if we believe we can avoid the situations they protect against.

This does mean that we will also have the Merkle contract sit behind an upgradeable proxy.

This PR also makes some small changes across the codebase to reduce binary size, as we are once again beyond the limit. The biggest lift here was defining the `process_party` helper for the `process_match_settle` method on the darkpool, but other changes include using statically-sized arrays in various serialization code, and avoiding error type conversions when we're going to unwrap anyway.

Finally, we make the `DarkpoolContract` methods generic over a type `S: TopLevelStorage + BorrowMut<Self>`, making it possible for the `DarkpoolContract` to make multiple calls to other contracts in one external method when it is a parent contract (as is the case when the `DarkpoolTestContract` inherits from it). Much of the diff in this PR is largely cosmetic, stemming from this change.

Tests for the new access controls are not yet written in order to not overload this review.

---

## EDIT

After discussion, we decided to `delegatecall` the Merkle contract after all, as this simplifies the smart contract topology significantly, and it is simple to avoid calling `selfdestruct` in any implementations of the Merkle contract (a doc comment has been left for this).

As a result, the only change in terms of access controls in this PR is to make the darkpool `Initializable`, we don't need to worry about `Ownable` for the Merkle contract.

Also, I am somewhat confused as to how the current implementation doesn't lead to storage collisions, but it doesn't seem to. I would think that the Merkle contract's fields would overwrite those in the Darkpool, since [Stylus lays out storage variables in the same way as the EVM](https://docs.arbitrum.io/stylus/reference/rust-sdk-guide#sol_storage:~:text=The%20types%20in%20%23%5Bsolidity_storage%5D%20are%20laid%20out%20in%20the%20EVM%20state%20trie%20exactly%20as%20they%20are%20in%20Solidity.), but ad hoc tests seem to show that everything continues working as expected. I'll save this as a potential problem to fully suss out later.

### DIFF
--- a/contracts-core/src/crypto/merkle.rs
+++ b/contracts-core/src/crypto/merkle.rs
@@ -1,6 +1,5 @@
 //! A sparse Merkle tree implementation, intended to be used in a smart contract context.
 
-use alloc::{vec, vec::Vec};
 use ark_ff::Zero;
 use common::{serde_def_types::ScalarFieldDef, types::ScalarField};
 use serde::{Deserialize, Serialize};
@@ -34,6 +33,7 @@ where
 
 /// Represents a node in the Merkle tree,
 /// including the height and index "coordinates" along with the value
+#[derive(Clone, Copy, Default)]
 pub struct NodeMetadata {
     pub height: usize,
     pub index: u128,
@@ -71,12 +71,20 @@ where
 
     /// Insert a value into the Merkle tree,
     /// returning the updated internal nodes from the insertion
-    pub fn insert(&mut self, value: ScalarField) -> Vec<NodeMetadata> {
+    pub fn insert(&mut self, value: ScalarField) -> [NodeMetadata; HEIGHT] {
         assert!(self.next_index < 2_u128.pow((HEIGHT - 1) as u32));
-        let node_changes = insert_helper(self, value, HEIGHT - 1, self.next_index, true);
-        self.root = node_changes.first().unwrap().value;
+        let mut updated_nodes = [NodeMetadata::default(); HEIGHT];
+        insert_helper(
+            self,
+            value,
+            HEIGHT - 1,
+            self.next_index,
+            true,
+            &mut updated_nodes,
+        );
+        self.root = updated_nodes[0].value;
         self.next_index += 1;
-        node_changes
+        updated_nodes
     }
 }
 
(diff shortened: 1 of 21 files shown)


contracts-core/src/crypto/merkle.rs
```diff
@@ -71,12 +71,20 @@ where
...
-        let node_changes = insert_helper(self, value, HEIGHT - 1, self.next_index, true);
-        self.root = node_changes.first().unwrap().value;
+        let mut updated_nodes = [NodeMetadata::default(); HEIGHT];
+        insert_helper(
+            self,
+            value,
+            HEIGHT - 1,
+            self.next_index,
+            true,
+            &mut updated_nodes,
```
joeykraut, line 83: what's the concrete reason for passing this in? If it's efficiency that's fine but this interface is a bit clunkier and departs from what we do elsewhere in our codebase.

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -55,66 +60,105 @@ pub struct DarkpoolContract {
...
+        // Set the caller as the owner
+        this.ownable.transfer_ownership(msg::sender()).unwrap();
+
+        // Set the verifier & Merkle addresses
+        this.verifier_address.set(verifier_address);
+        this.merkle_address.set(merkle_address);
+
+        // Mark the darkpool as initialized
+        this.initializable._initialize(1);
+
```
joeykraut, line 89: let's emit an event here, these things are helpful for looking up metadata on-chain

contracts-stylus/src/utils/helpers.rs
```diff
@@ -15,9 +17,25 @@ pub fn serialize_statement_for_verification<S: ScalarSerializable>(
...
+    allow(dead_code)
+)]
+pub fn delegate_call_helper<C: SolCall>(
+    storage: &mut impl TopLevelStorage,
+    address: Address,
+    args: <C::Arguments<'_> as SolType>::RustType,
+) -> C::Return {
+    let calldata = C::new(args).encode();
+    let res = unsafe { delegate_call(storage, address, &calldata).unwrap() };
+    C::decode_returns(&res, true /* valudate */).unwrap()
```
joeykraut: nit: typo "valudate" -> "validate"

joeykraut (approved):

contracts-core/src/crypto/merkle.rs
akirillo: Yup, efficiency in terms of binary size, actually, the vector push methods we were using before took a surprising bit of space. The Merkle contract should fit without this anyway, I just saw the chance and took it.

PR #134, opened 2023-11-12
contracts-stylus: emit events

akirillo:
This PR adds event emission to all of the contracts. This mirrors the way events were emitted in the Cairo implementation, with a few differences:
1. Events for `Ownable` and `Initializable` are copied over from the OpenZeppelin implementations ([ref](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/v5.0.0))
2. Events for upgrades are not yet implemented as the upgradeable proxy is not yet implemented
3. We add `WalletCreated` and `MatchSettled` events, and these, along with the existing `WalletUpdated` event, also log the serialization of the public wallet shares (for indexing & recovery)

Tests asserting well-formedness of these events are deferred for later, perhaps until indexing functionality is implemented in the relayer. Additionally, these changes push the darkpool contract past the binary size limit once again, the mitigation of which is left for the next PR.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -29,7 +29,7 @@ use crate::utils::{
         VALID_WALLET_CREATE_CIRCUIT_ID, VALID_WALLET_UPDATE_CIRCUIT_ID,
     },
     helpers::{delegate_call_helper, serialize_statement_for_verification},
-    interfaces::{initCall, insertSharesCommitmentCall, rootCall, rootInHistoryCall, IERC20},
+    solidity::{initCall, insertSharesCommitmentCall, rootCall, rootInHistoryCall, IERC20},
 };
 
 use super::components::{initializable::Initializable, ownable::Ownable};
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -229,15 +234,21 @@ impl DarkpoolContract {
...
         );
 
-        // TODO: Emit wallet updated event w/ wallet blinder share
+        let wallet_blinder_share_hash = keccak(wallet_blinder_share);
+        let public_wallet_shares =
+            serialize_wallet_shares(&valid_wallet_create_statement.public_wallet_shares);
+        evm::log(WalletCreated {
+            wallet_blinder_share: wallet_blinder_share_hash.into(),
+            public_wallet_shares,
+        });
```
joeykraut: Why the share hash here? Why not the blinder share directly? Also we can discuss this as an optimization later, but we don't need to add the public wallet shares to the event. It's nice if possible, not sure the pricing model for event data though

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: See the comment in the `EVENTS` section of the `sol!` macro in `utils::solidity` module. Indexed `bytes` event parameters need to be hashed, this is part of the Solidity spec.

Good call re: excluding public wallet shares, we should be able to get tx hash -> calldata to get public wallet shares from the log. We pay 8 gas per byte of log data, which is half of what we'd pay for calldata, so not negligible.

I'll probably add this optimization now

akirillo: Oh, actually, I'd assume the gas costs for events data are priced in L2 gas, but regardless, will implement this now - also because it helps chip away at binary size 😅

PR #135, opened 2023-11-13
contracts-stylus: optimize binary size after introduction of events & components

akirillo:
This PR makes the necessary changes to make the darkpool test contract deployable onchain. Primarily, this meant:
1. Removing the `components` module and inlining the `Ownable` & `Initializable` component functionality into the darkpool contract, removing unnecessary functions along the way
    - Apart from the unnecessary functions / events, this incurred overhead through Stylus SDK for initializing the `Ownable` / `Initializable` structs as [`StorageType`s](https://docs.rs/stylus-sdk/latest/stylus_sdk/storage/trait.StorageType.html)
2. Removing non-relayer-critical events
    - The only remaining events are those that relayers will need to actively listen for, i.e. new verification keys, external transfers, and new wallet shares being posted. Other helpful events about contract configuration, e.g. contract initialization, ownership transfer, etc. are removed _for now_.
3. Inlining implementations of `ScalarSerializable` for types other than the high-level statement types.
    - The contract / testing code only calls `serialize_to_scalars` on the high-level statement types. The other types from which statements are composed only implement `ScalarSerializable` for convenience in the implementation of `ScalarSerializable` for the statement types. This incurred overhead in terms of unnecessary vector operations, result handling, and monomorphization.

All existing tests pass. Tests asserting well-formedness of the events are deferred for the next PR for brevity.


common/src/custom_serde.rs
```diff
@@ -279,16 +219,22 @@ impl ScalarSerializable for ValidWalletUpdateStatement {
...
+            let external_transfer = self.external_transfer.as_ref().unwrap();
+            scalars.push(address_to_scalar(external_transfer.account_addr)?);
+            scalars.push(address_to_scalar(external_transfer.mint)?);
+            scalars.extend(u256_to_scalars(external_transfer.amount)?);
+            scalars.push(external_transfer.is_withdrawal.into());
+        } else {
+            // If no external transfer, we extend with the serialization of a default ExternalTransfer,
+            // which is 5 zero-scalars:
+            // 1 for the account address, 1 for the mint, 2 for the amount, 1 for the is_withdrawal flag
+            scalars.extend(vec![ScalarField::zero(); 5]);
```
joeykraut: This is brittle, what happens when we add/remove a field from the struct? A better way to do this might be to serialize an `ExternalTransfer::default`.

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -66,43 +63,69 @@ pub struct DarkpoolContract {
...
 
-        // Set the caller as the owner
-        this.ownable.transfer_ownership(msg::sender()).unwrap();
+        // Set the verifier & Merkle addresses
+        this.verifier_address.set(verifier_address);
+        this.merkle_address.set(merkle_address);
 
         // Mark the darkpool as initialized
-        this.initializable._initialize(1);
+        DarkpoolContract::_initialize(storage, 1);
```
joeykraut, line 113: I'm a fan of these changes from a simplicity standpoint as well :)

contracts-stylus/src/contracts/test_contracts/darkpool_test_contract.rs
```diff
@@ -16,36 +13,10 @@ struct DarkpoolTestContract {
...
-impl Borrow<Initializable> for DarkpoolTestContract {
-    fn borrow(&self) -> &Initializable {
-        &self.darkpool.initializable
-    }
-}
-
 // Expose internal helper methods of the Darkpool contract used in testing
 #[external]
-#[inherit(DarkpoolContract, Ownable, Initializable)]
+// #[inherit(DarkpoolContract, Ownable, Initializable)]
```
joeykraut: dead code

joeykraut (approved):

common/src/custom_serde.rs
akirillo: yeah, this is just so concise... but i'll give that a go and see if the binary fits

akirillo: hmm, to be fair - if we add/remove a field from the struct, we'll have to change the hand-written serialization code above, as well, so the blast radius doesn't change much

joeykraut: Sources of brittle code should not justify each other, these are just bugs waiting to happen

akirillo: ok, well said - i think i found a nice middle ground here via a helper function for serializing external transfers, and use that on the existing transfer or a `ExternalTransfer::default`, and it fits (not only that, but actually made the binary another 200 bytes smaller than before!)

PR #136, opened 2023-11-14
deploy-scripts: rudimentary deploy script for TransparentUpgradeableProxy

akirillo:
This PR initializes the `deploy-scripts` crate with a rudimentary deploy script for a [`TransparentUpgradeableProxy`](https://docs.openzeppelin.com/contracts/5.x/api/proxy#transparent_proxy). This script assumes that the proxy ABI & bytecode is compiled beforehand, and expects file paths pointing to these artifacts as CLI args.

Down the line, the functionality currently implemented in the `integration` crate for initializing the darkpool contract will be moved into here, when we make deploy scripts for the Stylus smart contracts.

Tests asserting proper functionality of the upgradeable proxy are deferred for the next PR, but at a minimum, ad-hoc testing shows that the proxy & proxy admin contracts deploy successfully.


deploy-scripts/src/cli.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(Args)]
+pub struct DeployProxyArgs {
+    /// Path to the proxy contract ABI
+    #[arg(short, long)]
+    pub abi: String,
+
+    /// Path to the proxy contract bytecode
+    #[arg(short, long)]
+    pub bytecode: String,
```
joeykraut: why not just hardcode these or have them pull & compile from a known implementation? Reduces the surface area for easy bugs and simplifies the CLI. For something as static and critical as the UUPS proxy (realistically may never change) this is a defensive measure, but not a limiting one.

deploy-scripts/src/commands.rs
```diff
@@ -0,0 +1,92 @@
...
+            )
+            .await
+            .map_err(|_| DeployError::ContractInteraction)?
+            [NUM_BYTES_STORAGE_SLOT - NUM_BYTES_ADDRESS..NUM_BYTES_STORAGE_SLOT],
+    );
+
+    // TODO: Set up better logging
+    println!("Proxy contract deployed at {:#x}", proxy_address);
+    println!(
+        "Proxy admin contract deployed at {:#x}",
```
joeykraut, line 87: nit: just "Proxy admin: ...", we didn't deploy an admin and this could be misleading

deploy-scripts/src/commands.rs
```diff
@@ -0,0 +1,92 @@
...
+            .get_storage_at(
+                proxy_address,
+                // Can `unwrap` here since we know the storage slot constitutes a valid H256
+                H256::from_str(PROXY_ADMIN_STORAGE_SLOT).unwrap(),
+                None, /* block */
+            )
+            .await
+            .map_err(|_| DeployError::ContractInteraction)?
+            [NUM_BYTES_STORAGE_SLOT - NUM_BYTES_ADDRESS..NUM_BYTES_STORAGE_SLOT],
+    );
```
joeykraut, line 82: Isn't this just the `owner_address` we just passed in? Why not just print that address? Esp because this code is pretty hard to read (i.e. without contextualizing on the storage layout of OZ proxy).

As a user of the CLI, it doesn't tell me anything different to print `owner_address`, I assume that the script correctly setup the admin. If we're worried about bugs, we should test it to cover our bases, no need to actually check a storage slot.

joeykraut (approved):
just a few small things, i.e. deferring to tests

PR #141, opened 2023-11-20
contracts-core: merkle: use correct empty leaf value

akirillo:
This PR adds a constant representing the empty leaf value used by the relayer, and makes sure that the merkle tree is initialized using this constant at the leaf layer. Tests are updated to reflect this and pass

### DIFF
--- a/common/src/constants.rs
+++ b/common/src/constants.rs
@@ -1,5 +1,11 @@
 //! Constants that parameterize the Plonk proof system
 
+use core::marker::PhantomData;
+
+use ark_ff::{BigInt, Fp};
+
+use crate::types::ScalarField;
+
 /// The number of wire types in the circuit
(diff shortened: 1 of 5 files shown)


common/src/constants.rs
```diff
@@ -46,3 +52,14 @@ pub const MERKLE_HEIGHT: usize = 32;
...
+/// The value of an empty leaf in the Merkle tree
+pub const EMPTY_LEAF_VALUE: ScalarField = Fp(
+    BigInt([
+        3955535776564458318,
+        6962397718615243800,
+        1102221708206984009,
+        1286929648744757623,
+    ]),
+    PhantomData,
+);
```
joeykraut: Same as in the corresponding relayer PR, let's add a doc comment to disambiguate this nothing-up-my-sleeve

joeykraut (approved):

PR #142, opened 2023-11-27
scripts: add vkey upload script

akirillo:
This PR introduces a new script for uploading verification keys, using the circuit definitions from the relayer. Ad-hoc testing of the script shows that it works (i.e., the transaction is executed), though ad-hoc integration tests with the relayer's Arbitrum client don't have on-chain verification succeeding. Fixing that is out of scope for this PR.

### DIFF
--- a/scripts/src/cli.rs
+++ b/scripts/src/cli.rs
@@ -9,8 +9,8 @@ use clap::{Args, Parser, Subcommand, ValueEnum};
 use ethers::providers::Middleware;
 
 use crate::{
-    commands::{build_and_deploy_stylus_contract, deploy_proxy, upgrade},
-    errors::DeployError,
+    commands::{build_and_deploy_stylus_contract, deploy_proxy, upgrade, upload_vkey},
+    errors::ScriptError,
 };
 
 #[derive(Parser)]
@@ -33,6 +33,7 @@ pub enum Command {
     DeployProxy(DeployProxyArgs),
     DeployStylus(DeployStylusArgs),
     Upgrade(UpgradeArgs),
+    UploadVkey(UploadVkeyArgs),
 }
 
 impl Command {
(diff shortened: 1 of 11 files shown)


scripts/src/cli.rs
```diff
@@ -125,3 +130,28 @@ pub struct UpgradeArgs {
...
+    #[arg(short, long)]
+    pub circuit: Circuit,
+
+    /// The address of the darkpool proxy contract
+    #[arg(short, long)]
+    pub darkpool_address: String,
+
+    /// Whether or not to use the testing circuit size parameters
+    #[arg(short, long)]
+    pub test: bool,
```
joeykraut: maybe `small` instead, test implies that the command is being run as a test, small suggests the actual effect of this flag

scripts/src/utils.rs
```diff
@@ -175,3 +191,73 @@ pub fn deploy_stylus_contract(
...
+            .map_err(|_| ScriptError::ConversionError)?,
+        q_comms: try_unwrap_commitments(&jf_vkey.selector_comms)?,
+        sigma_comms: try_unwrap_commitments(&jf_vkey.sigma_comms)?,
+        g: jf_vkey.open_key.g,
+        h: jf_vkey.open_key.h,
+        x_h: jf_vkey.open_key.beta_h,
+    })
+}
+
+pub fn get_vkey_bytes(circuit: Circuit, test: bool) -> Result<Vec<u8>, ScriptError> {
```
joeykraut: Perhaps `vkey_to_bytes`, `get` implies this is some field of the verifier key

scripts/src/utils.rs
```diff
@@ -175,3 +191,73 @@ pub fn deploy_stylus_contract(
...
+) -> Result<[G1Affine; N], ScriptError> {
+    comms
+        .iter()
+        .map(|c| c.0)
+        .collect::<Vec<_>>()
+        .try_into()
+        .map_err(|_| ScriptError::ConversionError)
+}
+
+pub fn convert_jf_vkey(jf_vkey: VerifyingKey<Bn254>) -> Result<VerificationKey, ScriptError> {
```
joeykraut, line 209: This should have a doc comment or be renamed. Consider the perspective of someone who is reading about our system for the first time, what is `jf`, what conversion is going on here, is it runtime type? These things are obvious to us but won't be to a first time reader

joeykraut (approved):

scripts/src/utils.rs
akirillo: Ack, I'll do `gen_vkey_bytes`, don't want the method name to imply that a vkey should be passed in

PR #143, opened 2023-11-28
contracts-core: merkle: make height exclusive of root

akirillo:
This PR changes the Merkle tree implementation to have its height parameter be exclusive of the root. This is consistent with how Merkle openings / authentication paths use the height parameter in the relayer, making integration testing far more straightforward.

Unit tests for Merkle implementation pass, as do relayer-side integration tests for the Merkle contract.


joeykraut (approved):
Wise change

PR #144, opened 2023-11-29
contracts-stylus, scripts: add no-verify feature for darkpool

akirillo:
This PR adds a `no-verify` feature to the darkpool which disables proof & ECDSA verification. This is intended to be used when testing the Arbitrum client in the relayer, as proper verification logic (and thus proper inputs to verification) are out of scope for the client tests and make them quite cumbersome.

Additionally, this PR adds a flag to the `scripts` CLI for toggling the `no-verify` feature when building/deploying contracts.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -275,12 +275,16 @@ impl DarkpoolContract {
             valid_wallet_update_statement.merkle_root,
         );
 
-        assert!(ecdsa_verify::<StylusHasher, PrecompileEcRecoverBackend>(
-            &valid_wallet_update_statement.old_pk_root,
-            valid_wallet_update_statement_bytes.as_slice(),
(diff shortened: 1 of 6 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -237,16 +237,21 @@ impl DarkpoolContract {
...
-            .into();
-
-        assert!(DarkpoolContract::verify(
-            storage,
-            VALID_WALLET_CREATE_CIRCUIT_ID,
-            proof,
-            public_inputs
-        ));
+        // Note: the "no-verify" feature is ONLY for testing purposes
+        #[cfg(not(feature = "no-verify"))]
```
joeykraut: might be cleaner to encapsulate this conditional compilation to a helper method like `verify_proof` or something. That way we don't have to mark every instance with the conditional block

joeykraut (approved):

PR #145, opened 2023-11-29
meta, contracts-stylus: darkpool: bump stylus sdk to 0.4.2 & add storage gap

akirillo:
This PR upgrades the Stylus SDK to version `0.4.2`, which includes a fix ensuring that delegate calls are properly issued when using the `delegatecall` helper function from the Stylus SDK. As a result, delegate calls to the Merkle contract started causing storage collisions with the darkpool contract, so a storage gap was added to the darkpool contract to avoid this.

Tests pass.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -20,14 +20,17 @@ use stylus_sdk::{
     crypto::keccak,
     evm, msg,
     prelude::*,
-    storage::{StorageAddress, StorageBool, StorageBytes, StorageMap, StorageU64},
+    storage::{
+        StorageAddress, StorageArray, StorageBool, StorageBytes, StorageMap, StorageU256,
+        StorageU64,
+    },
 };
 
 use crate::utils::{
     backends::{PrecompileEcRecoverBackend, StylusHasher},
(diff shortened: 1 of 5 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -39,6 +39,10 @@ use crate::utils::{
 #[solidity_storage]
 #[cfg_attr(feature = "darkpool", entrypoint)]
 pub struct DarkpoolContract {
+
+    /// Storage gap to prevent collisions with the Merkle contract
+    __gap: StorageArray<StorageU256, 64>,
```
joeykraut: Can we either constantize this `64` with a readable name or explain why 64 exactly in the doc comment? It's not obvious to smart contract newbies like myself

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: sure, will constantize, but it's completely arbitrary. it's sort of like a budget on the amount of flexibility we want to have with changes to the merkle contract state. if the state grows, it may eat into this 64-slot gap

PR #147, opened 2023-11-30
contracts-stylus: utils: define assert macro conditional on no-verify feature

akirillo:
This PR changes the way the `no-verify` feature is implemented by replacing all previous conditional compilation blocks with the usage of a `assert_if_verifying` macro that wraps the underlying `assert!` macro in a conditional compilation block that is active when the `no-verify` feature is unset. This allows us to cleanly avoid asserting valid historical roots, and double-spent nullifiers for the client integration tests.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -26,16 +26,20 @@ use stylus_sdk::{
     },
 };
 
-use crate::utils::{
-    backends::{PrecompileEcRecoverBackend, StylusHasher},
-    constants::{
-        STORAGE_GAP_SIZE, VALID_COMMITMENTS_CIRCUIT_ID, VALID_MATCH_SETTLE_CIRCUIT_ID,
-        VALID_REBLIND_CIRCUIT_ID, VALID_WALLET_CREATE_CIRCUIT_ID, VALID_WALLET_UPDATE_CIRCUIT_ID,
-    },
-    helpers::{delegate_call_helper, keccak_hash_scalar, serialize_statement_for_verification},
-    solidity::{
-        initCall, insertSharesCommitmentCall, rootCall, rootInHistoryCall,
-        ExternalTransfer as ExternalTransferEvent, VerificationKeySet, WalletUpdated, IERC20,
+use crate::{
+    assert_if_verifying,
+    utils::{
+        backends::{PrecompileEcRecoverBackend, StylusHasher},
+        constants::{
+            STORAGE_GAP_SIZE, VALID_COMMITMENTS_CIRCUIT_ID, VALID_MATCH_SETTLE_CIRCUIT_ID,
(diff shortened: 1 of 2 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -493,24 +493,15 @@ impl DarkpoolContract {
...
-            let this = storage.borrow_mut();
-            let vkey_bytes = this.verification_keys.get(circuit_id).get_bytes();
-            assert!(!vkey_bytes.is_empty());
-
-            let verifier_address = this.verifier_address.get();
-            let verification_bundle_ser = [vkey_bytes, proof.into(), public_inputs.into()].concat();
-            let result = static_call(storage, verifier_address, &verification_bundle_ser).unwrap();
+        let this = storage.borrow_mut();
+        let vkey_bytes = this.verification_keys.get(circuit_id).get_bytes();
+        assert_if_verifying!(!vkey_bytes.is_empty());
```
joeykraut: do we need this? Can this method even be called if the verifier is disabled?

contracts-stylus/src/utils/helpers.rs
```diff
@@ -55,3 +55,21 @@ pub fn delegate_call_helper<C: SolCall>(
...
+        }
+    };
+    ($cond:expr, $($arg:tt)+) => {
+        // Note: the "no-verify" feature is ONLY for testing purposes
+        #[cfg(not(feature = "no-verify"))]
+        {
+            assert!($cond, $($arg)+);
+        }
+    };
+}
```
joeykraut, line 75: ITT this can be achieved in a single branch:
```rust
#[macro_export]
macro_rules! assert_if_verifying {
    ($cond:expr $(, $($arg:tt)*)?) => {
        // Note: the "no-verify" feature is ONLY for testing purposes
        #[cfg(not(feature = "no-verify"))]
        {
            assert!($cond $(, $($arg)*)?);
        }
    };
}
```

joeykraut (approved):

contracts-stylus/src/utils/helpers.rs
akirillo: I just copied the branches of the original assert macro, would rather defer to that for consistency

contracts-stylus/src/contracts/darkpool.rs
akirillo: True, we don't need to use the macro here, my rationale was just to reduce cognitive load on the reader by using this everywhere... but as I think about I don't think that's a good precedent, so I'll remove the usage here

contracts-stylus/src/utils/helpers.rs
joeykraut: Consistency with a standard lib impl? Why not just make the code simpler?

akirillo: i can just turn my brain off and know for a fact that it will function exactly like the std lib, lol - but if it's confusing to read, i can change it

PR #153, opened 2023-12-08
scripts, integration: reading/writing srs instead of vkeys

akirillo:
This PR replaces the `gen-vkeys` script with a `gen-srs` script. This because we need to have consistent verification *and* proving keys between deployment/testing. Additionally, the integration tests were updated to use the saved SRS.

All integration tests pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,13 +34,10 @@ eyre = "0.6.8"
 ethers = "2.0"
 clap = { version = "4.4.7", features = ["derive"] }
 tokio = { version = "1.12.0", features = ["full"] }
-renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/circuits-refactor", default-features = false }
-circuits = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/circuits-refactor", features = [
-    "test_helpers",
-] }
-circuit-types = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/circuits-refactor", features = [
-    "test-helpers",
-] }
+constants = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+circuits = { git = "https://github.com/renegade-fi/renegade.git" }
+circuit-types = { git = "https://github.com/renegade-fi/renegade.git" }
 itertools = "0.12"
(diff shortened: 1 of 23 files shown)


Cargo.toml
```diff
@@ -34,11 +34,12 @@ eyre = "0.6.8"
 ethers = "2.0"
 clap = { version = "4.4.7", features = ["derive"] }
 tokio = { version = "1.12.0", features = ["full"] }
-renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/circuits-refactor", default-features = false }
-circuits = { git = "https://github.com/renegade-fi/renegade.git", branch = "joey/circuits-refactor", features = [
+constants = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+circuits = { git = "https://github.com/renegade-fi/renegade.git", features = [
     "test_helpers",
 ] }
```
joeykraut: Why do we need the `test_helpers` feature enabled? It should probably just be enabled in `dev-dependencies` to avoid using unsafe methods

scripts/src/cli.rs
```diff
@@ -126,21 +129,20 @@ pub struct UpgradeArgs {
...
     pub calldata: Option<String>,
-
-    /// Whether or not to use the darkpool test contract address
-    /// as the new implementation address
-    #[arg(short, long)]
-    pub test: bool,
+    // /// Whether or not to use the darkpool test contract address
+    // /// as the new implementation address
+    // #[arg(short, long)]
+    // pub test: bool,
```
joeykraut: Is this intended to be commented out? If so why not just remove it

scripts/src/commands.rs
```diff
@@ -179,26 +170,12 @@ pub async fn upgrade(
...
-            gen_vkey(Circuit::ValidCommitments)?,
-            gen_vkey(Circuit::ValidReblind)?,
-            gen_vkey(Circuit::ValidMatchSettle)?,
-        ]
-    };
+pub fn gen_srs(args: GenSrsArgs) -> Result<(), ScriptError> {
+    let mut rng = thread_rng();
 
-    write_vkeys(&args.vkeys_path, vkeys, args.test)?;
+    // Generate universal SRS
```
joeykraut, line 176: Since we're using the unsafe universal setup method let's log a warning here indicating that this is unsafe and shouldn't be used in production

joeykraut (approved):

PR #154, opened 2023-12-12
contracts-stylus: merkle: use merkle struct fields directly in storage

akirillo:
This PR removes the native Rust `SparseMerkleTree` type and inlines it's fields & methods into the `MerkleContract` implementation. This makes our storage access patterns significantly more efficient - we no longer de/serialize the entire `SparseMerkleTree` to/from storage on every insertion.

**Improvements:**
Storage access during insertion would previously cost us an aggregate ~5,673,266 gas => ~$1.26 (assuming $2200 ETH).
As a result of the changes in this PR, these costs are significantly reduced. The storage access patterns taking place in the insertion method consume 210,043 gas, and the overall cost of Merkle insertion went down from ~$2.67 -> ~$1.45, a ~46% improvement.

For what it's worth, this also shaved the Merkle binary down from 20.2kb -> 15.6kb.

**Testing:**
Integration tests checking Merkle functionality (`merkle`, `new-wallet`, `update-wallet`, `process-match-settle`) all pass.


joeykraut (approved):
great work

PR #157, opened 2023-12-13
contracts-core: verifier: use batch inversion for public input polynomial

akirillo:
This PR changes the implementation of the verifier to use Montgomery batch inversion when calculating the Lagrange bases used to evaluate the public inputs polynomial.

**Improvements:**
Whereas the parts of verification that relied on the Lagrange bases (steps 6 and 7) previously consumed 2,504,087 gas ($0.55), they now (including the batch inversion procedure) consume 438,181 gas ($0.1), a reduction of 2,065,906 gas ($0.45).
Overall, the cost of `update_wallet` went from 10,853,503 gas ($2.39) to 8,743,558 gas ($1.92), a 19.5% improvement.

**Testing:**
Verifier unit tests & integration tests invoking the verifier contract all pass.


contracts-core/src/verifier/mod.rs
```diff
@@ -140,35 +139,34 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
         let mut pi_eval = lagrange_1_eval * public_inputs[0];
-        for (i, pi) in public_inputs.iter().enumerate().skip(1) {
-            let omega_i = domain.element(i);
-            let lagrange_i_eval = zero_poly_eval_div_n * omega_i / (*zeta - omega_i);
-            pi_eval += lagrange_i_eval * pi;
+        for ((o_i, l_i), p_i) in domain_elements
+            .iter()
+            .zip(lagrange_bases.iter())
+            .zip(public_inputs.iter())
+            .skip(1)
```
joeykraut, line 167: you can use [`izip`](https://docs.rs/itertools/latest/itertools/macro.izip.html) for this, i believe `itertools` is already in scope

joeykraut (approved):

contracts-core/src/verifier/mod.rs
akirillo: we don't have `itertools` as a dependency here :/ so i'd rather not add one so as mitigate binary bloat

PR #158, opened 2023-12-15
contracts-core: transcript: custom hash to scalar to allow inlined arkworks for verifier

akirillo:
This PR replaces the invocation of `ScalarField::from_le_bytes_mod_order` in the `transcript` module with a custom implementation that replicates the functionality while avoiding the `ark-serialize` code path previously incurred.

This decreases the binary size of the verifier enough to allow for disabling the `bin-opt` feature on the `ark-ff` patch, enabling inlining and loop unrolling in the modular arithmetic implementations.

As a result, the cost of verification in `update_wallet` goes down from 1,842,783 gas to 1,572,002 gas, about a 15% improvement.

Transcript & verification unit & integration tests pass.

### DIFF
--- a/contracts-core/src/transcript/mod.rs
+++ b/contracts-core/src/transcript/mod.rs
@@ -6,8 +6,8 @@ use alloc::vec::Vec;
 use ark_ff::PrimeField;
 use common::{
     backends::HashBackend,
-    constants::TRANSCRIPT_STATE_SIZE,
-    custom_serde::{BytesSerializable, TranscriptG1},
+    constants::{HASH_SAMPLE_BYTES, NUM_BYTES_FELT, TRANSCRIPT_STATE_SIZE},
+    custom_serde::{bigint_from_le_bytes, BytesSerializable, TranscriptG1},
     types::{Challenges, G1Affine, Proof, ScalarField, VerificationKey},
 };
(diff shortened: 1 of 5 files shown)


contracts-core/src/transcript/mod.rs
```diff
@@ -47,7 +47,27 @@ impl<H: HashBackend> Transcript<H> {
...
         self.state.copy_from_slice(&[buf0, buf1].concat());
 
-        ScalarField::from_le_bytes_mod_order(&self.state[..48])
+        // Sample the first 48 bytes of hash output into a scalar.
+        // Follows the implementation of `PrimeField::from_le_bytes_mod_order`
+
+        // We begin by taking the highest `NUM_BYTES_FELT-1` bytes of the hash output in little-endian order
+        // and converting them into a scalar directly, as no reduction is needed.
+        let (remaining_bytes, bytes_to_directly_convert) =
+            self.state[..48].split_at(48 - (NUM_BYTES_FELT - 1));
```
joeykraut: Can we constantize the `48` here? Might help give semantic meaning to the number

joeykraut (approved):

contracts-core/src/transcript/mod.rs
akirillo: ack

PR #160, opened 2023-12-16
contracts-stylus: verification keys contract

akirillo:
This PR changes the implementation of the Darkpool contract to no longer use constantized verification keys. Rather, we deploy a separate contract which holds these, from which the darkpool queries verification keys via `staticcall`. This reduces the darkpool's binary bloat enough to allow for 1) more aggressive compiler optimizations, and 2) removal of the arkworks patch.

Unfortunately, these changes are almost exactly net 0 in cost (the optimizations cover the overhead of `staticcall`ing the vkeys contract). However, the removal of the Arkworks fork and the extra flexibility in the Darkpool contract are worth it, IMO.

Integration tests pass.


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -170,29 +178,32 @@ impl DarkpoolContract {
...
+                static_call_helper::<validWalletUpdateCall>(storage, vkeys_address, ()).into();
+
+            assert!(DarkpoolContract::verify(
+                storage,
+                valid_wallet_update_vkey_bytes,
+                proof,
+                serialize_statement_for_verification(&valid_wallet_update_statement)
+                    .unwrap()
+                    .into(),
+            ));
```
joeykraut, line 205: Unsure what the complexity of this is, but we could consider refactoring these three steps into a single helper for consistency across other `if_verifying` blocks

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -397,28 +414,38 @@ impl DarkpoolContract {
...
-
-        assert_if_verifying!(DarkpoolContract::verify(
-            storage,
-            VALID_REBLIND_VKEY_BYTES,
-            valid_reblind_proof,
-            serialize_statement_for_verification(&match_payload.valid_reblind_statement)
-                .unwrap()
-                .into()
-        ));
+        if_verifying!({
```
joeykraut, line 417: this method will get called twice right? once for each party, probably worth moving the logic to load the vkeys into the caller to avoid doubling the `static_call`s

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: yes, this is done in a later PR

contracts-stylus/src/contracts/darkpool.rs
akirillo: yes, this is done in a later PR

contracts-stylus/src/contracts/darkpool.rs
akirillo: I tried doing this before, it becomes quite complex to satisfy the trait bounds around the generic `SolCall`'s (e.g. here, `validWalletUpdateCall`) argument & return types. Ends up looking a lot hairier IMO

PR #162, opened 2023-12-17
contracts-core: transcript: less mod math when squeezing challenges

akirillo:
This PR changes the implementation of `get_and_append_challenge` in the `transcript` module to use less modular arithmetic when sampling a scalar from the hash output. Previously, this would invoke 25 modular multiplications and additions, it now invokes only one of each.

**Improvements:**
As a result of this, the cost of `process_match_settle` decreased from 11,358,713 gas to 11,096,155 gas, an approximate $0.06 cost reduction (now at a total cost of $2.44). Note, of course, that challenges are computed 5 times in `process_match_settle`, so one can extrapolate that this saves us about 52,511.6 gas per verification.

**Testing:**
Transcript generation unit tests, and integration tests invoking verification all pass.

### DIFF
--- a/contracts-core/src/transcript/mod.rs
+++ b/contracts-core/src/transcript/mod.rs
@@ -3,10 +3,10 @@
 pub mod errors;
 
 use alloc::vec::Vec;
-use ark_ff::PrimeField;
+use ark_ff::{BigInt, BigInteger, PrimeField};
 use common::{
     backends::HashBackend,
-    constants::{HASH_SAMPLE_BYTES, NUM_BYTES_FELT, TRANSCRIPT_STATE_SIZE},
+    constants::{HASH_SAMPLE_BYTES, SPLIT_INDEX, TRANSCRIPT_STATE_SIZE},
     custom_serde::{bigint_from_le_bytes, BytesSerializable, TranscriptG1},
     types::{Challenges, G1Affine, Proof, ScalarField, VerificationKey},
 };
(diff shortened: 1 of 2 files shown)


contracts-core/src/transcript/mod.rs
```diff
@@ -48,26 +48,30 @@ impl<H: HashBackend> Transcript<H> {
...
-        // Follows the implementation of `PrimeField::from_le_bytes_mod_order`
 
-        // We begin by taking the highest `NUM_BYTES_FELT-1` bytes of the hash output in little-endian order
+        // We begin by taking the lowest `NUM_BYTES_FELT-1` bytes of the hash output in little-endian order
         // and converting them into a scalar directly, as no reduction is needed.
-        let (remaining_bytes, bytes_to_directly_convert) =
-            self.state[..HASH_SAMPLE_BYTES].split_at(HASH_SAMPLE_BYTES - (NUM_BYTES_FELT - 1));
-        let mut res =
+        let (bytes_to_directly_convert, remaining_bytes) =
+            self.state[..HASH_SAMPLE_BYTES].split_at(NUM_BYTES_FELT - 1);
```
joeykraut: I'd add `const SPLIT_INDEX: usize = NUM_BYTES_FELT - 1`, it would help the readability of the sizing math below

contracts-core/src/transcript/mod.rs
```diff
@@ -48,26 +48,30 @@ impl<H: HashBackend> Transcript<H> {
...
-            res *= window_size;
-            res += ScalarField::from(*byte);
-        }
-        res
+        // Next, we interpret the remaining bytes in little-endian order as a scalar.
+        // Again, no reduction is needed.
+        let mut rem_scalar =
+            ScalarField::from_bigint(bigint_from_le_bytes(remaining_bytes).unwrap()).unwrap();
+
+        // Now, we shift the latter scalar left by 31 bytes, which is equivalent to multiplying by 2^249.
```
joeykraut: This doc comment does not quite reflect what the code does, we multiply by $$2^{249}$$ but this is not equivalent to shifting by 31 bytes, can we reword this comment and some of the naming in variables

joeykraut (approved):

contracts-core/src/transcript/mod.rs
akirillo: The doc comment was actually incorrect - we multiply by 2^248, which is indeed a shift of 31 bytes

PR #167, opened 2023-12-19
contracts-core: verifier: batch verification

akirillo:
This PR implements batch verification in the verifier as it is done in [Jellyfish](https://github.com/renegade-fi/mpc-jellyfish/blob/main/plonk/src/proof_system/verifier.rs#L199), namely by extending the final pairing check to be over a random linear combination of the G1 elements each verification instance produces for its respective final pairing check.

Additionally, this extends the Jellyfish implementation by batch-inverting the Lagrange bases across all verification instances in the batch.

**Testing:**
Unit tests for verification pass. While the verifier contract has been adapted to this interface, it is not yet tested - this will be done in the subsequent PR.

### DIFF
--- a/contracts-core/src/verifier/mod.rs
+++ b/contracts-core/src/verifier/mod.rs
@@ -6,7 +6,7 @@
 pub mod errors;
 
 use alloc::{vec, vec::Vec};
-use ark_ff::{batch_inversion_and_mul, FftField, Field, One, Zero};
+use ark_ff::{batch_inversion, FftField, Field, One, Zero};
 use common::{
     backends::{G1ArithmeticBackend, HashBackend},
     constants::NUM_WIRE_TYPES,
@@ -50,9 +50,10 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
 
         let num_proofs = proofs.len();
 
-        let mut g1_lhs_elems = Vec::with_capacity(num_proofs);
-        let mut g1_rhs_elems = Vec::with_capacity(num_proofs);
-        let mut final_challenges = Vec::with_capacity(num_proofs);
+        let mut challenges_batch = Vec::with_capacity(num_proofs);
+        let mut zero_poly_evals_batch = Vec::with_capacity(num_proofs);
+        let mut domain_elements_batch = Vec::with_capacity(num_proofs);
+        let mut lagrange_bases_batch = Vec::with_capacity(num_proofs);
 
         for ((vkey, proof), public_inputs) in
             vkeys.iter().zip(proofs.iter()).zip(public_inputs.iter())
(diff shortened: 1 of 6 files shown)


contracts-core/src/verifier/mod.rs
```diff
@@ -413,29 +440,29 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
 
-    /// Batch validate all evaluations
-    fn step_12(
-        &mut self,
+    /// Compute G1 elements to be used in the final pairing check
+    /// for the given proof.
+    ///
+    /// This is the final G1 arithmetic done in step 12 of the verifier algorithm
+    /// before the pairing check.
+    fn step_12_part_1(
```
akirillo, line 496: I'm not sure how I feel about `step_12_part_1` and `step_12_part_2` as function names, since by this point, we've deviated from the original Plonk paper. I can rename these two functions to be more representative of what they accomplish, but that might mean also doing so for the rest of the `step_*` functions...

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
 };
 use core::{marker::PhantomData, result::Result};
 
-use crate::transcript::Transcript;
+use crate::transcript::{serialize_scalars_for_transcript, Transcript};
 
 use self::errors::VerifierError;
 
-/// The verifier struct, which encapsulates the verification key, transcript, and backend for elliptic curve arithmetic.
+/// The verifier struct, which is colored by the backends used for elliptic curve arithmetic and hashing.
```
joeykraut: Can we change the wording here? Maybe "defined generically over EC arithmetic and hashing backends", "colored" is ambiguous, esp in rust terms

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
 
-        self.step_3(public_inputs, &vkey)?;
+        let mut challenges_batch = Vec::with_capacity(num_proofs);
+        let mut zero_poly_evals_batch = Vec::with_capacity(num_proofs);
+        let mut domain_elements_batch = Vec::with_capacity(num_proofs);
+        let mut lagrange_bases_batch = Vec::with_capacity(num_proofs);
 
-        let challenges = self.step_4(&vkey, proof, public_inputs, extra_transcript_init_message)?;
+        for ((vkey, proof), public_inputs) in
+            vkeys.iter().zip(proofs.iter()).zip(public_inputs.iter())
```
joeykraut: You can use [`izip`](https://docs.rs/itertools/latest/itertools/macro.izip.html) to make this iterator cleaner

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
+                ScalarField::get_root_of_unity(vkey.n).ok_or(VerifierError::InvalidInputs)?;
+
+            let zero_poly_eval = Self::step_5(domain_size, &challenges);
+
+            // Precompute Lagrange bases (zeta^n - 1)/(n*(zeta - omega_i)) using Montgomery's batch inversion trick
+            let mut domain_elements: Vec<ScalarField> = Vec::with_capacity(public_inputs.len());
+            domain_elements.push(ScalarField::one());
+            for i in 0..public_inputs.len() - 1 {
+                domain_elements.push(domain_elements[i] * omega);
+            }
```
joeykraut: A suggested optimization, these can certainly be re-used across proofs. I.e. we do:
1. Let `m = max(vkeys[0].n, ..., vkeys[k].n)`
2. Compute the evaluation domain for a circuit of size `m.next_power_of_two()`
3. For each circuit thereafter, we can subsample the correctly sized evaluation domain from this larger one, noting that the `n`th roots of unity also contain all the `n/2`th roots of unity

Certainly this optimization is only worth it at some cost profile, and should probably be in a follow-up if we choose to take it.

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
-            self.vkey.n
-        } else {
-            self.vkey
-                .n
-                .checked_next_power_of_two()
-                .ok_or(VerifierError::InvalidInputs)?
-        };
-        let omega =
-            ScalarField::get_root_of_unity(self.vkey.n).ok_or(VerifierError::InvalidInputs)?;
+        batch_inversion(&mut lagrange_bases_batch);
```
joeykraut: I'd break everything above this point into a helper method, this function is quite long

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
-        let mut domain_elements: Vec<ScalarField> = Vec::with_capacity(public_inputs.len());
-        domain_elements.push(ScalarField::one());
-        for i in 0..public_inputs.len() - 1 {
-            domain_elements.push(domain_elements[i] * omega);
-        }
+        for (i, ((vkey, proof), public_inputs)) in vkeys
+            .iter()
+            .zip(proofs.iter())
+            .zip(public_inputs.iter())
+            .enumerate()
```
joeykraut: `izip` here as well would be simpler

contracts-core/src/verifier/mod.rs
```diff
@@ -5,115 +5,164 @@
...
+        {
+            let challenges = &challenges_batch[i];
+            let zero_poly_eval = zero_poly_evals_batch[i];
+            let domain_elements = &domain_elements_batch[i];
+            let lagrange_bases: Vec<ScalarField> = lagrange_bases_batch
+                [lagrange_bases_cursor..lagrange_bases_cursor + public_inputs.len()]
+                .iter()
+                .map(|l| l * &zero_poly_eval)
+                .collect();
```
joeykraut: I'd break this up into two statements for simplicity, with a good variable name in between

contracts-core/src/verifier/mod.rs
joeykraut: I think the current naming is fine, from the reader's perspective it makes most sense this way

contracts-core/src/verifier/mod.rs
```diff
@@ -445,19 +472,63 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
+    /// Compute the final pairing check for a batch of proofs.
+    ///
+    /// For the verification of a single proof, we do a pairing check of the form:
+    /// e(A, [x]2) == e(B, [1]2)
+    ///
+    /// Now, for batch verification over `m` proofs, we extend the pairing check to the following:
+    /// e(A0 + ... + r^{m-1} * Am, [x]2) = e(B0 + ... + r^{m-1} * Bm, [1]2)
+    ///
+    /// By the Schwartz-Zippel lemma, for a random `r`, this check will succeed with overwhelming
+    /// probability if and only if the individual pairing checks do.
```
joeykraut, line 536: Can we add a code link and properly attribute `jellyfish` here?

joeykraut (approved):
lgtm overall, have some suggestions to improve the readability, given the system-critical nature of this code

contracts-core/src/verifier/mod.rs
akirillo: I'd rather not include itertools as a dependency to mitigate binary size

contracts-core/src/verifier/mod.rs
akirillo: I err on the side of minimizing binary size as above

contracts-core/src/verifier/mod.rs
akirillo: I like this idea, and the cost should be pretty tame, I think - all w have to do is calculate the m-th root of unity and maintain a potentially-growing list of sampled domain elements as we go through each proof.

I'll leave this for a follow-up

contracts-core/src/verifier/mod.rs
akirillo: I err on the side of minimizing binary size as above

contracts-core/src/verifier/mod.rs
akirillo: Took a stab at this, also bringing in the batch inversion changes in #168. The function is still, deliberately, a bit long, as I want it to be read as a progression through the steps. That said, I've taken a lot of the auxiliary logic that is not directly part of the steps out into helpers

PR #168, opened 2023-12-20
contracts-stylus: darkpool: adapt to batch verification interface

akirillo:
This PR changes the darkpool contract to reflect the batch verification process. This means that verification keys, proofs, and public inputs passed to the verifier contract are expected to be contained in vectors representing the batches. This changes some assumptions about the calldata serialization for external darkpool methods, e.g. the proof submitted to `new_wallet` is expected to be the serialization of a single-element vector containing the proof, and the vkeys contract returns the serializations of vectors containing the verification keys, etc. These changes are all documented in code comments.

**Improvements:**
As a result of these changes, `process_match_settle` takes 9,202,379 gas, costing us $2.02. `update_wallet` takes 3,832,898 gas, costing $0.84, i.e. it is unchanged (technically, a bit cheaper)

**Testing:**
Integration tests invoking verification pass. Arbitrum client integration tests in the relayer repo may pass due to verification being disabled, but the client has not been updated to reflect these interface changes.


contracts-core/src/verifier/mod.rs
```diff
@@ -42,21 +42,25 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
         let mut domain_elements_batch = Vec::with_capacity(num_proofs);
-        let mut lagrange_bases_batch = Vec::with_capacity(num_proofs);
+        let mut all_lagrange_bases = Vec::with_capacity(num_proofs);
 
-        for ((vkey, proof), public_inputs) in
-            vkeys.iter().zip(proofs.iter()).zip(public_inputs.iter())
+        for ((vkey, proof), public_inputs) in vkey_batch
+            .iter()
+            .zip(proof_batch.iter())
+            .zip(public_inputs_batch.iter())
```
joeykraut: I think that this is in #167 as well, but `izip` would be nice for simplicity here, same below as well

PR #171, opened 2023-12-22
contracts-stylus: verifier: `verify` & `verify_match_settle` methods on verifier contract

akirillo:
This PR changes the verifier contract from a minimal stylus entrypoint (no ABI, one external method) to a full contract exposing separate methods for single-proof verification, and verification of the `process_match_settle` proofs (in an application-specific way, i.e. expecting the exact number of verification keys / proofs / public inputs).

The application-specific structure here lets us avoid some considerable serialization overhead, the impact of which is outlined below.

As a part of this PR, in order to get both the verifier and the darkpool to fit within the binary size limit, I had to make some miscellaneous changes. In the verifier, this meant replacing a lot of iterator methods with for loops + indexing, and in the darkpool, this meant changing the vkeys contract to return the `VALID COMMITMENTS`, `VALID REBLIND`, & `VALID MATCH SETTLE` verification keys in a single function call, which is also more efficient due to less static call overhead.

**Improvements:**
As a result, `process_match_settle` now takes 9,174,764 gas, i.e. $2.02. `update_wallet` comes in at $0.84, a slight improvement.

**Testing:**
`new_wallet`, `update_wallet`, `process_match_settle`, & `verifier` tests all pass.


Cargo.toml
```diff
@@ -57,7 +57,7 @@ incremental = false      # no incremental builds
 inherits = "dev"
 codegen-units = 1   # prefer efficiency to compile time
 panic = "abort"     # use simple panics
-opt-level = "z"     # optimize for size ("s" may also work)
+opt-level = 3       # optimize for size ("s" may also work)
```
joeykraut: comment here is out of date

contracts-stylus/src/contracts/verifier.rs
```diff
@@ -1,51 +1,55 @@
...
-        .collect::<Vec<_>>();
-
-    let mut verifier = Verifier::<PrecompileG1ArithmeticBackend, StylusHasher>::default();
-
-    let result = verifier
-        .verify(&vkey_batch, &proof_batch, &public_inputs_batch)
-        .unwrap();
-
-    Ok(vec![result as u8])
+#[external]
```
joeykraut, line 15: super clean

joeykraut (approved):

PR #174, opened 2023-12-28
contracts-core: verifier: scaffold type & interface changes for proof linking

akirillo:
This PR makes the interface changes necessary for proof linking, namely:
1. Create a `LinkingProof` struct containing the information necessary for proving correct input linking, and a `LinkingVerificationKey` struct containing preprocessed information used for verifying a linking proof
2. Break out separate `verify` and application-specific `verify_match_bundle` methods on the `Verifier` struct - this makes downstream implementation of proof linking much more straightforward than in the generic case

**Testing:**
Unit tests for verification pass


common/src/types.rs
```diff
@@ -24,12 +24,14 @@ pub type MontFp256<P> = Fp256<MontBackend<P, NUM_U64S_FELT>>;
...
 #[serde_as]
-#[derive(Clone, Copy, Serialize, Deserialize)]
+#[derive(Clone, Copy, Serialize, Deserialize, Debug)]
 pub struct VerificationKey {
     /// The number of gates in the circuit
     pub n: u64,
     /// The number of public inputs to the circuit
     pub l: u64,
+    /// The number of linked inputs in the circuit
+    pub num_linked_inputs: u64,
```
joeykraut: A general point, we don't just have one domain linked per proof. How does this capture that one to many relationship?

Perhaps an argument for moving out of the verification key, but up to you

common/src/types.rs
```diff
@@ -63,12 +65,18 @@ pub struct Proof {
     /// The commitments to the split quotient polynomials
     #[serde_as(as = "[G1AffineDef; NUM_WIRE_TYPES]")]
     pub quotient_comms: [G1Affine; NUM_WIRE_TYPES],
+    /// The commitment to the linking quotient polynomial
+    #[serde_as(as = "G1AffineDef")]
+    pub linking_quotient_comm: G1Affine,
```
joeykraut: Same meta point here: some proofs will not have a link some will have many. This doesn't seem to capture that.

More broadly I think we should remove this field from the `Proof` structure regardless of the decision to include/remove parameterizations in the vkey. The proof here is of a _different_ protocol, and we should highlight that with a different type. I think it's okay to add `commitments_reblind_link_proof`-esque arguments to the interface. Verbosity isn't a dirty interface on it's own!

Doing so also provides a way to very simply represent the one-to-many relationship between proofs and links.

contracts-core/src/verifier/mod.rs
```diff
@@ -34,17 +34,69 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Default for Verifier<G, H> {
...
+        proof: &Proof,
+        public_inputs: &PublicInputs,
+    ) -> Result<bool, VerifierError> {
+        // Steps 1 & 2 of the verifier algorithm are assumed to be completed by this point,
+        // by virtue of the type system. I.e., the proof should be deserialized in a manner such that
+        // elements not in the scalar field, and points not in G1, would cause a panic.
+
+        Self::step_3(public_inputs, vkey)?;
+
+        let challenges = Self::step_4(vkey, proof, public_inputs)?;
```
joeykraut, line 54: this doesn't look like it actually changed, let me know if it needs review.

common/src/types.rs
joeykraut: We also need some way to represent the *indices* at which the proof is linked, it does not begin at zero

contracts-core/src/verifier/mod.rs
akirillo: Correct, doesn't need review

common/src/constants.rs
```diff
@@ -12,6 +12,11 @@ pub const NUM_WIRE_TYPES: usize = 5;
 /// The number of selectors in the circuit
 pub const NUM_SELECTORS: usize = 13;
 
+/// The number of proofs in a match bundle
+pub const NUM_MATCH_PROOFS: usize = 5;
+/// The number of circuits in a match bundle
+pub const NUM_CIRCUITS: usize = 3;
```
joeykraut, line 18: nit: maybe `NUM_MATCH_CIRCUITS`, the current name gives a sense of a more general meaning

common/src/types.rs
```diff
@@ -50,9 +50,50 @@ pub struct VerificationKey {
     pub x_h: G2Affine,
 }
 
+/// The Plonk verification keys used when verifying the matching of a trade
```
joeykraut: nit: 
```suggestion
/// The Plonk verification keys used when verifying the match and settlement of a trade
```

common/src/types.rs
```diff
@@ -50,9 +50,50 @@ pub struct VerificationKey {
...
+    /// The number of linked inputs, equivalently the size of the subdomain
+    link_group_size: usize,
+}
+
+/// The linking verification keys used when verifying the matching of a trade
+#[derive(Serialize, Deserialize, Default)]
+pub struct MatchLinkingVkeys {
+    /// The verification key for the
+    /// `PARTY 0 VALID COMMITMENTS` <-> `PARTY 0 VALID REBLIND` link
+    pub party_0_valid_commitments_valid_reblind_linking_vkey: LinkingVerificationKey,
```
joeykraut: this is not going to be different between party 0 and party 1, the only one that is different is the link into `VALID MATCH SETTLE` of the two parties' `VALID COMMITMENTS` proof

we should also shorten these names.

common/src/types.rs
```diff
@@ -80,6 +121,51 @@ pub struct Proof {
...
+    pub linking_poly_opening: G1Affine,
+}
+
+/// The linking proofs used to ensure input consistency
+/// between the `MatchProofs`
+#[derive(Serialize, Deserialize, Default)]
+pub struct MatchLinkingProofs {
+    /// The proof of linked inputs between
+    /// `PARTY 0 VALID COMMITMENTS` <-> `PARTY 0 VALID REBLIND`
+    pub party_0_valid_commitments_valid_reblind_linking_proof: LinkingProof,
```
joeykraut: nit: i'd shorten these names too, it's pretty verbose, maybe `valid_reblind_commitments0`? The fact that it's a link proof is implied by its container

joeykraut (approved):
a few nits, though the abstractions look really good to me

common/src/constants.rs
akirillo: i make this change in a later PR haha

common/src/types.rs
akirillo: Yup, I dedupe in a later PR, lol
But ack on shortening names

PR #175, opened 2023-12-30
contracts-core: verifier: batch-verify linking proofs in `verify_match`

akirillo:
This PR implements the verification algorithm for linking proofs, and batch-verifies all of the linking proofs in a match alongside the associated Plonk proofs.

No testing has been done yet, so as to keep this PR shorter and focused on the verification design.

### DIFF
--- a/contracts-core/src/verifier/mod.rs
+++ b/contracts-core/src/verifier/mod.rs
@@ -91,7 +91,7 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
         let (lhs_g1, rhs_g1) =
             Self::step_12_part_1(f_1, neg_e_1, domain_elements[1], proof, &challenges)?;
 
-        Self::step_12_part_2(&[lhs_g1], &[rhs_g1], &[challenges.u], vkey.x_h, vkey.h)
+        Self::batch_opening(&[lhs_g1], &[rhs_g1], &[challenges.u], vkey.x_h, vkey.h)
     }
 
     /// Batch-verifies:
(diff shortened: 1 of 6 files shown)


contracts-core/src/verifier/mod.rs
```diff
@@ -227,6 +259,128 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
+        let mut subdomain = Vec::with_capacity(link_group_size);
+        subdomain.push(link_group_generator.pow([link_group_offset as u64]));
+        for i in 0..link_group_size - 1 {
+            subdomain.push(subdomain[i] * link_group_generator);
+        }
+
+        let mut zero_poly_eval = ScalarField::one();
+        for omega_i in subdomain {
+            zero_poly_eval *= eta - omega_i
+        }
```
joeykraut: We do not need to materialize the subdomain into a vector, we can directly compute the vanishing polynomial's evaluation without first allocating

contracts-core/src/verifier/mod.rs
```diff
@@ -227,6 +259,128 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
+        )?;
+
+        // Prepare LHS & RHS G1 elements for pairing check
+        let g1_lhs = linking_poly_opening;
+        let g1_rhs = G::msm(
+            &[eta, ScalarField::one(), -ScalarField::one()],
+            &[
+                linking_poly_opening,
+                linking_poly_comm,
+                G1Affine::identity(),
```
joeykraut: We don't need to add the third term here, if `G` is the bn254 identity element, then `-1 * G = G` and `G + G'` for any `G'` is equal to `G'`.

contracts-core/src/verifier/mod.rs
```diff
@@ -385,7 +531,7 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
             Self::step_9_line_4(zero_poly_eval, proof, challenges)?,
         ];
 
-        G::msm(&[ScalarField::one(); 4], &points).map_err(|_| VerifierError::ArithmeticBackend)
+        Ok(G::msm(&[ScalarField::one(); 4], &points)?)
```
joeykraut: nit: Just map the error with `Into::into`, code is messier when wrapping like this, same elsewhere

joeykraut (approved):

PR #176, opened 2024-01-02
contracts-core: verifier: proof linking unit tests

akirillo:
This PR adds unit tests for proof linking, both in the single and batch case. Additionally, it breaks out Plonk proof processing logic the overarching `verify_match` method to make existing unit tests for the Plonk proofs in a match bundle more straightforwards & appropriately scoped.

All unit tests in the `verifier` module pass.


contracts-core/src/transcript/mod.rs
```diff
@@ -128,14 +124,27 @@ impl<H: HashBackend> Transcript<H> {
...
             gamma,
             alpha,
             zeta,
             v,
             u,
-        })
+        }
+    }
+
+    pub fn compute_linking_proof_challenge(
```
joeykraut, line 141: nit: this method could use a doc comment, to an unfamiliar reader it is not obvious that these challenges are for a different polynomial protocol

contracts-core/src/verifier/mod.rs
```diff
@@ -829,6 +858,294 @@ mod tests {
...
+        let mut evaluations = domain.fft(&poly.coeffs);
+        evaluations.splice(
+            link_group_offset..link_group_offset + link_group.len(),
+            link_group,
+        );
+        let coeffs = domain.ifft(&evaluations);
+        poly.coeffs = coeffs;
+    }
+
+    fn add_shared_link_group<R: CryptoRng + RngCore>(
```
joeykraut, line 945: we should replace these methods with the prover side utility when that is complete. the link group domains are a bit more complicated than splicing evaluations, especially when the polys are of different native domains. Not worth replicating all that logic here, so we can just use prover side utils

Can we add a todo comment for that?

contracts-core/src/verifier/mod.rs
```diff
@@ -829,6 +858,294 @@ mod tests {
...
+        );
+
+        let party_0_valid_commitments_valid_match_settle_linking_vkey = LinkingVerificationKey {
+            link_group_generator: domain.group_gen,
+            link_group_offset: party_0_valid_commitments_valid_match_settle_offset,
+            link_group_size: LINK_GROUP_SIZE,
+        };
+
+        // Prove the `PARTY 1 VALID COMMITMENTS` <-> `VALID MATCH SETTLE` linking
+        let (party_1_valid_commitments_valid_match_settle_linking_proof, _, _) = prove_link(
```
joeykraut: nit: I'd shorten these names

contracts-core/src/verifier/mod.rs
```diff
@@ -829,6 +858,294 @@ mod tests {
...
+        poly_2: &mut DensePolynomial<ScalarField>,
+        domain: &impl EvaluationDomain<ScalarField>,
+        link_group_offset: usize,
+    ) {
+        let link_group = random_scalars(LINK_GROUP_SIZE, rng);
+        add_link_group_to_poly(poly_1, domain, link_group_offset, link_group.clone());
+        add_link_group_to_poly(poly_2, domain, link_group_offset, link_group);
+    }
+
+    fn compute_single_link_test_data<R: CryptoRng + RngCore>(
```
joeykraut, line 962: nit: Can we add a doc comment that also explains the use of the `link_polys` flag?

joeykraut (approved):

contracts-core/src/verifier/mod.rs
akirillo: yes absolutely

PR #177, opened 2024-01-02
contracts-stylus: darkpool: expect signature over wallet shares

akirillo:
This PR changes the implementation of `update_wallet` to expect the `public_inputs_signature` to be over the big-endian representation of `poseidon_hash(private_shares_commitment || public_wallet_shares)`, as opposed to over the entire serialized `ValidWalletUpdateStatement`. This way, we expect the user to sign _exactly_ the state that is committed to in the Merkle tree. This required moving the ECDSA verification logic into the Merkle contract (where we compute said Poseidon hash).

**Testing:**
The `update_wallet` test passes, all other code paths are untouched.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -2,14 +2,10 @@
 //! verifying the various proofs of the Renegade protocol, and handling deposits / withdrawals.
 
 use alloc::{vec, vec::Vec};
-use common::{
-    custom_serde::BytesSerializable,
-    types::{
-        ExternalTransfer, MatchPayload, ScalarField, ValidMatchSettleStatement,
-        ValidWalletCreateStatement, ValidWalletUpdateStatement,
-    },
+use common::types::{
+    ExternalTransfer, MatchPayload, PublicSigningKey, ScalarField, ValidMatchSettleStatement,
+    ValidWalletCreateStatement, ValidWalletUpdateStatement,
 };
-use contracts_core::crypto::ecdsa::ecdsa_verify;
 use core::borrow::{Borrow, BorrowMut};
 use stylus_sdk::{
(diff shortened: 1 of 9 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -338,19 +344,41 @@ impl DarkpoolContract {
...
+        let merkle_address = storage.borrow_mut().merkle_address.get();
+
+        let old_pk_root_u256s = old_pk_root
+            .x
+            .into_iter()
+            .chain(old_pk_root.y)
+            .map(scalar_to_u256)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
```
joeykraut: nit: might read nicer if this was broken out into a helper

contracts-stylus/src/contracts/merkle.rs
```diff
@@ -89,12 +98,44 @@ where
...
-            .collect();
+        let shares_commitment = self.compute_shares_commitment(shares);
+
+        self.insert_helper(shares_commitment, P::HEIGHT as u8, insert_index, true);
+
+        Ok(())
+    }
+
+    /// Computes a commitment to the given wallet shares & inserts it into the Merkle tree
+    pub fn verify_state_sig_and_insert(
```
joeykraut: nit: Would be good to add a comment somewhere in this file explaining that we do sig verification in Merkle to avoid hashing in the main contract, for binary size reasons. That way we know the assumptions that, when removed, will allow us to move sig verification into the main contract

joeykraut (approved):

contracts-stylus/src/contracts/merkle.rs
akirillo: well put, will do so

PR #178, opened 2024-01-05
docs: specification: initial spec draft

akirillo:
This PR introduces a draft specification of the contracts. This currently was mostly an exercise for helping me think through protocol assumptions and invariants for further testing / auditing, but I find it useful to commit to the repo for future posterity & orientation in the codebase.

The contextual information at the start of the doc can be polished later, the pertinent portion up for review here is the assumptions / invariants.

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -27,7 +27,7 @@ These changes are encapsulated in a [Plonk](https://eprint.iacr.org/2019/953.pdf
 
 If the change was the settlement of a matched trade, then the `DarkpoolContract` also verifies "proof linking" between the Plonk proofs involved - this is a cryptographic statement attesting that the Plonk proofs use the same public inputs where appropriate. This, too is, done via a `staticcall` to the `VerifierContract` (and also fetches preprocessed information from the `VkeysContract`).
 
-The updated wallet is committed to as a leaf in a global Merkle tree via a `delegatecall` to the `MerkleContract` from the `DarkpoolContract`. The `MerkleContract` also maintains a set of historic roots, to which it adds the newly-computed root.
+The updated wallet is committed to as a leaf in a global Merkle tree via a `delegatecall` to the `MerkleContract` from the `DarkpoolContract`. We also add the newly-computed root to a set of historic roots.
 
 If the change was an update to an existing wallet by the user (this excludes the settlement of a matched trade, which can be done by a relayer), we also verify an ECDSA signature by the user over the commitment to the updated wallet to be inserted into the Merkle tree.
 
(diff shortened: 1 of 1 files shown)


docs/specification.md
```diff
@@ -0,0 +1,172 @@
...
+
+Users submit changes to their wallet (e.g., creating a new wallet, creating / cancelling orders, settling matched trades) to the `DarkpoolContract`.
+
+These changes are encapsulated in a [Plonk](https://eprint.iacr.org/2019/953.pdf) zero-knowledge proof(s), which the `DarkpoolContract` verifies by `staticcall`ing the `VerifierContract` (during this process, the `DarkpoolContract` also `staticcall`s the `VkeysContract` for preprocessed public information about the statement being proven).
+
+If the change was the settlement of a matched trade, then the `DarkpoolContract` also verifies "proof linking" between the Plonk proofs involved - this is a cryptographic statement attesting that the Plonk proofs use the same public inputs where appropriate. This, too is, done via a `staticcall` to the `VerifierContract` (and also fetches preprocessed information from the `VkeysContract`).
+
+The updated wallet is committed to as a leaf in a global Merkle tree via a `delegatecall` to the `MerkleContract` from the `DarkpoolContract`. The `MerkleContract` also maintains a set of historic roots, to which it adds the newly-computed root.
```
joeykraut: nit: Can we reword "The `MerkleContract` also maintains..." to something indicating that the `DarkpoolContract` (or if you like the proxy) holds this state? The `delegatecall` in the previous sentence implies this but I'd rather make it crystal clear that all state is maintained in the darkpool contract, behind access controls

docs/specification.md
```diff
@@ -0,0 +1,172 @@
...
+- `verify_match`
+    - Batch-verifies the passed-in Plonk & linking proofs together. This is only intended to be called by the `DarkpoolContract`'s `process_match_settle` method, in that it expects the same proofs as that method does. This, too, incurs many `staticcall`s to the `ecAdd` and `ecMul` precompiles, and a single `staticcall` to `ecPairing`
+
+## Assumptions
+
+We make the following assumptions about interaction with the protocol:
+- We deploy the contracts correctly (e.g. Merkle tree is correct height, verification keys are correct)
+- The `ProxyAdmin` is initialized with an owner account that we securely control
+- The `initialize` method of the `DarkpoolContract` is called in the deployment of the `TransparentUpgradeableProxy`
+- Management of a wallet is only delegated to a single relayer
```
joeykraut: This should be "cluster" not "relayer"

docs/specification.md
```diff
@@ -0,0 +1,172 @@
...
+- Calldata is generally available for wallet indexing & recovery
+- Wallet blinders & the keychain are properly derived from their respective hash chains (we don't prove this in-circuit)
+- Users will not authorize a malicious key rotation
```
joeykraut: I feel these last two are out of scope of the contracts, happy to discuss but these more fall under a broader theme of "inputs are well formed" which are validated elsewhere in the system (i.e. by the relayer)

docs/specification.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Protocol Invariants
+
+The following invariants, organized by contract for readability, must hold true in all possible states of the protocol.
+
+`DarkpoolContract`
```
joeykraut: I'd recommend stating these more in terms of security guarantees on _state transitions_. So for example we think of an entry in the Merkle tree as being a unit of state, and an invariant is that a state is only ever "spent" once. This statement is more closely related to the product level invariants we care about and captures (abstractly) the behavior of the nullifier set and its tie to the Merkle tree through our proofs. Too much focus on implementation isn't really an invariant, rather a description of implementation

joeykraut (approved):
Nice work, have a few nits around wording. particularly with the invariants so they are "extensible" beyond their implementation

PR #180, opened 2024-01-07
test-helpers: create dummy circuits using circuit-types traits

akirillo:
This PR begins the effort of refactoring the `test-helpers` crate, and other components of the testing suite, to be more ergonomic and extensible. Here, we replace the creation of dummy Renegade circuits (`VALID WALLET CREATE`, etc.) with the actual `SingleProverCircuit` trait from the relayer. This will make testing proof linking between the circuits significantly easier by leveraging the exact implementation relayer-side.

Further PRs will continue to de-duplicate utilities between this repo and the relayer repo, this diff had already grown quite large.

**Testing:**
Verification unit tests pass, but unfortunately, I'm not yet able to run the integration tests since the proof linking changes have made the verifier exceed the binary size limit. Mitigating this is out of scope for this PR.


Cargo.toml
```diff
@@ -61,3 +64,10 @@ opt-level = 3       # optimize for performance (default for most contracts)
 lto = true          # link time optimization
 rpath = false       # no run-time search path
 incremental = false # no incremental builds
+
+[patch."https://github.com/renegade-fi/renegade-contracts.git"]
+# The `arbitrum-client` crate, which we depend on in e.g. `test-helpers`,
+# depends on `contracts-common`, but from the git source.
+# In order to have consistent types in the `arbitrum-client` interfaces,
+# we patch the `contracts-common` dependency to point to the local crate.
+contracts-common = { path = "./contracts-common" }
```
joeykraut, line 73: this feels like an odd nearly circular dependency, might make more sense to refactor this dependency pattern. I don't feel strongly about this -- food for thought

test-helpers/src/dummy_renegade_circuits.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use crate::proof_system::apply_dummy_constraints;
+
+// TODO: Make a `prove_with_srs` method
+
+const DUMMY_DOMAIN_SIZE: usize = 1024;
+
+/// Generates a dummy circuit type named `Dummy$circuit` that implements the
+/// `SingleProverCircuit` trait using a dummy constraint set.
+macro_rules! generate_dummy_circuit {
```
joeykraut: I like this approach, but I think it could be simpler:

Seems like we introduce this macro to give circuits with different statement types. A suggestion that would simplify this is to use a single `DummyCircuit` with a generic for the statement type (no macro):

```rust
struct DummyCircuit<Statement: CircuitBaseType> { _phantom: ... }

impl<S: CircuitBaseType> SingleProverCircuit for DummyCircuit<S> {
    type Statement = S;
    type Witness = ();

    fn name() -> String { "dummy".to_string() }

    fn apply_constriants(...) { apply_dummy_constraints(...) }
}
```
Since really we only need to be generic over a type not a behavior. Thinking this might simplify things, macros are a pain!

test-helpers/src/proof_system.rs
```diff
@@ -34,20 +34,34 @@ use mpc_plonk::{
 use mpc_relation::{traits::Circuit, PlonkCircuit};
 use rand::thread_rng;
 
-use crate::misc::random_scalars;
+use crate::dummy_renegade_circuits::{
+    gen_process_match_settle_data, DummyValidCommitments, DummyValidMatchSettle, DummyValidReblind,
+    ProcessMatchSettleData,
+};
+
+pub fn apply_dummy_constraints(
```
joeykraut: Do we need any constraints at all? Not sure what adding constraints is testing at the end of the day, plus tests will run faster if we only use the constraints added by default to the cs

joeykraut (approved):
Nice work, very clean

Cargo.toml
akirillo: Yeah, it felt pretty hacky. Thinking about it now, I should be able to move in the conversion logic defined in `arbitrum-client` here, and import it there. I think that's all we use from it

test-helpers/src/dummy_renegade_circuits.rs
akirillo: Oi yes definitely

test-helpers/src/proof_system.rs
akirillo: Oh, I didn't realize there were default contraints, I guess for the proof-linking and public input gates? But regardless I'd like to have some constraints, for two reasons:
1. I do want some nontrivial arithmetic going on within the verifier, it functioning properly is in scope for these tests
2. This method is good scaffolding into which I'll later slot some basic link groups, which is definitely functionality I want to test

akirillo: hmm honestly facts

Cargo.toml
akirillo: Oh, hmm - I'm not sure that this patching approach is avoidable as long as we have types defined in different crates.

Currently, `arbitrum-client` imports `contracts-common` types and implements conversion, then `test-helpers` imports the conversion logic from `arbitrum-client` and the types from `contracts-common`.

I can't have the conversion logic defined in `contracts-common` due to WASMification issues, but even if I could (I could also define it in e.g. `test-helpers`), then we'd have the issue of `contracts-common` importing types from `circuits`, and `arbitrum-client` importing conversion logic from `contracts-common` and types from `circuits`.

I'm going to put off a cleaner solution to the type unification effort for down the line

PR #182, opened 2024-01-09
contracts-stylus: darkpool: add `Ownable` & `Pausable` controls

akirillo:
This PR makes the `DarkpoolContract` `Ownable` and `Pausable`. Now, all setter methods can be "paused" and "unpaused" by the contract owner, a role which can be transferred. Notably, we do not include events for ownership transfer or pausing, nor do we include an explicit `renounce_ownership` method. This is because these features take up binary size that is best left for other potential controls, and can be brought in via an upgrade once size constraints slacken.

**Testing:**
Tests have been added asserting proper `Ownable` and `Pausable` functionality, which pass (with the exception of asserting that `process_match_settle` works as intended when paused / unpaused, because completing the e2e tests for that is a separate workstream).


contracts-utils/src/conversion.rs
```diff
@@ -43,3 +48,27 @@ pub fn to_contract_vkey(
...
+            .y
+            .into_iter()
+            .map(Scalar::new)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap(),
+    };
+
+    CircuitPublicSigningKey { x, y }
+}
```
akirillo, line 74: this was moved, no need for review

contracts-utils/src/proof_system/test_data.rs
```diff
@@ -45,6 +53,58 @@ pub fn dummy_statement<R: RngCore + CryptoRng, S: CircuitBaseType>(rng: &mut R)
...
+        ]
+        .concat(),
+    );
+
+    let public_inputs_signature = Bytes::from(
+        hash_and_sign_message(&signing_key, &shares_commitment.serialize_to_bytes()).to_vec(),
+    );
+
+    Ok((proof, contract_statement, public_inputs_signature))
+}
```
akirillo, line 106: this was moved, no need for review

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -275,6 +335,32 @@ impl DarkpoolContract {
...
+    pub fn _check_owner<S: TopLevelStorage + Borrow<Self>>(storage: &S) {
+        assert_eq!(storage.borrow().owner.get(), msg::sender())
+    }
+
+    // ------------
+    // | PAUSABLE |
+    // ------------
+
+    /// Checks whether the darkpool's paused status matches the given `paused` value
+    pub fn _check_paused_status<S: TopLevelStorage + Borrow<Self>>(storage: &S, paused: bool) {
```
joeykraut: "check" here seems to imply that we'd return a bool, can we break this out into `_assert_paused` and `_assert_unpaused`? I assume those are reasonably small code size diffs, and that would make the actual enforced semantics much clearer at the callsite. Esp for something as critical as this.

integration/src/tests.rs
```diff
@@ -378,6 +379,165 @@ pub(crate) async fn test_initializable(
...
+        &contract_with_dummy_owner,
+        UNPAUSE_METHOD_NAME,
+        (),
+    )
+    .await?;
+
+    Ok(())
+}
+
+pub(crate) async fn test_pausable(
```
joeykraut, line 453: can we test somewhere in here that pausing is only allowed by the owner? might already be implicitly tested but if not that'd be a good base to cover

joeykraut (approved):
nicely done

integration/src/tests.rs
akirillo: yup, that's tested on lines 440 / 442

PR #183, opened 2024-01-09
contracts-stylus: darkpool: add protocol fee parameter

akirillo:
This PR adds a `protocol_fee` parameter to the `DarkpoolContract`, which is set during initialization and can be set by the owner.

Implementation details of the protocol fee (e.g., how many decimals are in the fixed-point representation) are deferred until it is implemented throughout the protocol. As such, fee-specific tests are also deferred.

**Testing:**
The ownership test was updated to assert that only the owner can set the protocol fee, and passes.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -61,7 +61,8 @@ pub struct DarkpoolContract {
     /// boolean indicating whether or not the nullifier is spent
     nullifier_set: StorageMap<U256, StorageBool>,
 
-    /// The protocol fee
+    /// The protocol fee, representing a percentage of the trade volume
+    /// as a fized-point number
     protocol_fee: StorageU256,
 }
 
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -60,6 +60,9 @@ pub struct DarkpoolContract {
     /// (which is a Bn254 scalar field element serialized into 32 bytes) to a
     /// boolean indicating whether or not the nullifier is spent
     nullifier_set: StorageMap<U256, StorageBool>,
+
+    /// The protocol fee
+    protocol_fee: StorageU256,
```
joeykraut, line 66: what's the representation here? This of course can happen later when we implement fees, but the relayer will be aware of a fee represented as a fixed point. Is the plan to have this U256 also represent a fixed point? If so can we call that out here?

scripts/src/cli.rs
```diff
@@ -76,6 +76,10 @@ pub struct DeployProxyArgs {
     /// and the underlying darkpool contract
     #[arg(short, long)]
     pub owner: String,
+
+    /// The initial protocol fee with which to initialize the darkpool contract
+    #[arg(short, long)]
+    pub fee: u64,
```
joeykraut, line 84: same question here wrt representation, u64 will probably fit any fee we'd like to make, but can we call that out here or more explicitly choose a repr?

joeykraut (approved):

PR #184, opened 2024-01-10
contracts-stylus: darkpool: add implementation address setter methods

akirillo:
This PR adds setter methods for the `VerifierContract`, `VkeysContract`, and `MerkleContract` implementation addresses stored in the `DarkpoolContract`.

This is the last of the necessary controls to add, so we can now think about which events will be the most useful to add (we won't be able to add events for all of the controls added).

**Testing:**
The ownership test was updated to assert that only the owner can set implementation addresses, and passes.
Additionally, I added a test for asserting that setting the implementation address does, indeed, point to a new contract. This test passes as well.

### DIFF
--- a/contracts-stylus/src/contracts/test_contracts/darkpool_test_contract.rs
+++ b/contracts-stylus/src/contracts/test_contracts/darkpool_test_contract.rs
@@ -1,14 +1,17 @@
 use core::borrow::BorrowMut;
 
 use alloc::vec::Vec;
-use contracts_common::types::ExternalTransfer;
+use contracts_common::{
+    constants::{MERKLE_ADDRESS_SELECTOR, VERIFIER_ADDRESS_SELECTOR, VKEYS_ADDRESS_SELECTOR},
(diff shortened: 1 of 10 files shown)


contracts-stylus/src/contracts/test_contracts/darkpool_test_contract.rs
```diff
@@ -36,6 +36,22 @@ impl DarkpoolTestContract {
         Ok(())
     }
 
+    pub fn is_implementation_upgraded(&mut self, address_selector: u8) -> Result<bool, Vec<u8>> {
+        let this = BorrowMut::<DarkpoolContract>::borrow_mut(self);
+        let implementation_address = match address_selector {
+            0 => this.verifier_address.get(),
+            1 => this.vkeys_address.get(),
+            2 => this.merkle_address.get(),
```
joeykraut: can we use the constants defined for these selectors?

joeykraut (approved):

PR #194, opened 2024-01-12
contracts-stylus, contracts-core: verifier: shrink verifier binary

akirillo:
This PR makes the necessary changes to bring the `VerifierContract` under the binary size limit. Concretely, these changes were:
1. Remove the separate single-proof-verification code path. This had some small efficiency gains (e.g., only had to do a single multiplication by the vanishing polynomial evaluation in the middle of batch inversion), but these did not justify the binary bloat.
2. Disable Stylus' storage cache feature for the `VerifierContract`. This feature was entirely foregone in the verifier anyway since it is stateless.

**Testing:**
All unit tests pass. Fixing the integration tests is deferred to a future PR.


contracts-core/src/verifier/mod.rs
```diff
@@ -44,54 +44,24 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
-        }
-
-        let f_1 = Self::step_10(d_1, &v_powers, vkey, proof)?;
-
-        let neg_e_1 = Self::step_11(r_0, &v_powers, vkey, proof, &challenges)?;
-
-        let (lhs_g1, rhs_g1) =
-            Self::step_12_part_1(f_1, neg_e_1, domain_elements[1], proof, &challenges)?;
+        // Prepare Plonk proofs for batch verification
+        let MatchOpeningElems {
```
joeykraut: nit: Isn't this a more general codepath than just for match-settle? If so probably better to generalize this name

contracts-core/src/verifier/mod.rs
```diff
@@ -44,54 +44,24 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
+            g1_lhs_elems,
+            g1_rhs_elems,
+            transcript_elements,
+        } = Self::prep_batch_plonk_proofs_opening(&[vkey], &[proof], &[public_inputs])?;
 
-        Self::batch_opening(&[lhs_g1], &[rhs_g1], &[challenges.u], vkey.x_h, vkey.h)
+        Self::batch_opening(
+            &g1_lhs_elems,
+            &g1_rhs_elems,
+            &transcript_elements,
```
joeykraut: Do we need to destructure the above opening elems? we just pass them right back in, may as well have `batch_opening` take in the `opening_elems`

joeykraut (approved):

PR #196, opened 2024-01-12
contracts-stylus: darkpool: restructure process_match_settle arguments

akirillo:
This PR tweaks the interface of the `process_match_settle` method on the `DarkpoolContract` to expect serializations of the more ergonomic `MatchProofs` and `MatchLinkingProofs` structs. These are what's expected by the verifier, as well, allowing them to be passed through without any re-serialization.

`process_match_settle` integration tests will be updated in the next PR.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -21,7 +21,8 @@ use crate::{
         constants::STORAGE_GAP_SIZE,
         helpers::{
             delegate_call_helper, pk_to_u256s, scalar_to_u256,
-            serialize_statement_for_verification, static_call_helper,
+            serialize_match_statements_for_verification, serialize_statement_for_verification,
+            static_call_helper,
         },
         solidity::{
             initCall, insertSharesCommitmentCall, processMatchSettleVkeysCall, rootCall,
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -336,21 +337,13 @@ impl DarkpoolContract {
...
-        party_1_valid_reblind_proof: Bytes,
-        valid_match_settle_proof: Bytes,
-        valid_match_settle_statement_bytes: Bytes,
-        _party_0_valid_commitments_valid_reblind_linking_proof: Bytes,
-        _party_0_valid_commitments_valid_match_settle_linking_proof: Bytes,
-        _party_1_valid_commitments_valid_reblind_linking_proof: Bytes,
-        _party_1_valid_commitments_valid_match_settle_linking_proof: Bytes,
+        valid_match_settle_statement: Bytes,
+        match_proofs: Bytes,
+        match_linking_proofs: Bytes,
```
joeykraut, line 350: clean, can we document the intended ordering of the proofs in the docstring here -- given that this is an external method?

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: yes, good call out

akirillo: I'll mention which structs it expects to be serialized, serde manages the ordering of the serialized fields etc consistently

PR #198, opened 2024-01-12
scripts, integration: deploy all contracts, run all tests

akirillo:
This PR updates the `scripts` and `integration` crates to support deploying all testing contracts, and running all tests, using a single command for each.

This includes some small changes to the utilities & how arguments are passed to tests which are entirely ergonomic, and change nothing about the functionality.

Finally, we make a small fix to `test_external_transfer` so that the dev account approves the darkpool to spend its funds.

**Testing:**
All unit & integration tests pass.

### DIFF
--- a/integration/src/abis.rs
+++ b/integration/src/abis.rs
@@ -48,7 +48,7 @@ abigen!(
     VerifierContract,
     r#"[
         function verify(bytes memory verification_bundle) external view returns (bool)
-    function verifyMatch(bytes memory match_bundle) external view returns (bool)
+        function verifyMatch(bytes memory match_bundle) external view returns (bool)
     ]"#
 );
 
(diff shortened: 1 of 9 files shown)


integration/src/abis.rs
```diff
@@ -67,6 +67,7 @@ abigen!(
     r#"[
         function mint(address memory _address, uint256 memory value) external
         function balanceOf(address memory _address) external view returns (uint256)
+        function approve(address memory spender, uint256 value) external returns (bool)
```
joeykraut: may as well just make it the full abi like we do in the [relayer repo](https://github.com/renegade-fi/renegade/blob/main/arbitrum-client/src/abi.rs#L53)

joeykraut (approved):

PR #201, opened 2024-01-16
contracts-stylus: verbose reverts

akirillo:
This PR removes all of the `assert` and `unwrap` invocations throughout contract code (covering the `contracts-stylus`, `contracts-core` and `contracts-common` crates) and replaces them with `Result` propagation so that top-level contract methods can return UTF-8 encoded error messages.

This included making some changes to compilation flags (unfortunately, had to bump the `DarkpoolContract` down to optimization level 2), and other tweaks to bring the verifier contract within the binary size limit.

**Testing:**
All unit and integration tests pass, and ad-hoc testing showed that errors are propagated correctly even across contract call boundaries.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,10 +1,3 @@
-[target.wasm32-unknown-unknown]
-rustflags = [
-  "-C", "link-arg=-zstack-size=32768",
-  "-C", "inline-threshold=0",
-  "-Zlocation-detail=none",
-]
-
 [target.aarch64-apple-darwin]
 rustflags = [
   "-C", "link-arg=-undefined",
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -60,7 +60,7 @@ incremental = false      # no incremental builds
 inherits = "dev"
 codegen-units = 1   # prefer efficiency to compile time
 panic = "abort"     # use simple panics
-opt-level = 3       # optimize for performance (default for most contracts)
+opt-level = "s"     # optimize for performance (default for most contracts)
```
joeykraut: nit: out of date comment

joeykraut (approved):

PR #202, opened 2024-01-17
scripts: erc20 deploy script

akirillo:
This PR adds a separate script which deploys instances of the dummy ERC20 contract for each of the provided tickers, and approves max spending of the darkpool for each of the provided accounts. A caveat here is that the provided tickers are not actually stored as the contract's `symbol`, this is due to the fact that Stylus doesn't support constructors and as such tickers are hardcoded into the contract code. This isn't a relevant feature right now, but down the line we could add a storage variable & setter method for the symbol if needed.

**Testing:**
The `external-transfer` test passes after these changes.


scripts/src/commands.rs
```diff
@@ -254,6 +261,70 @@ pub async fn deploy_proxy(
...
+
+            // Approve the darkpool to spend the maximum amount of tokens
+            erc20
+                .approve(darkpool_address, EthersU256::MAX)
+                .send()
+                .await
+                .map_err(|e| ScriptError::ContractInteraction(e.to_string()))?
+                .await
+                .map_err(|e| ScriptError::ContractInteraction(e.to_string()))?;
+        }
```
joeykraut: this may read easier if these are broken out into helpers -- the fn name will self-document

scripts/src/constants.rs
```diff
@@ -125,8 +125,13 @@ pub const VERIFIER_CONTRACT_KEY: &str = "verifier_contract";
 /// The vkeys contract key in the `deployments.json` file
 pub const VKEYS_CONTRACT_KEY: &str = "vkeys_contract";
 
-/// The dummy ERC20 contract key in the `deployments.json` file
-pub const DUMMY_ERC20_CONTRACT_KEY: &str = "dummy_erc20_contract";
+/// The ticker of the ERC20 contract deployed using `deploy_test_contracts`,
+/// which is also its contract key in the `deployments.json` file
```
joeykraut, line 129: is this still the case, won't the deployments.json entry be the ticker specified in the args, not the dummy ticker?

scripts/src/constants.rs
```diff
@@ -125,8 +125,13 @@ pub const VERIFIER_CONTRACT_KEY: &str = "verifier_contract";
...
+pub const DUMMY_ERC20_TICKER: &str = "DUMMY";
+
+/// The amount of dummy ERC20 tokens to fund the user with
+/// when deploying the testing contracts
+pub const TEST_FUNDING_AMOUNT: u64 = 1000;
```
joeykraut, line 134: Can we make this higher? Maybe something like 100,000. I'm thinking because if I'm testing an ETH/USDC trade this will not be enough usdc (generally)

joeykraut (approved):

scripts/src/constants.rs
akirillo: Oh, this is just the amount used in the `deploy_test_contracts` script which is only used for the integration tests in this repo, not in the integration tests in the relayer. There, we set the funding amount manually.

scripts/src/constants.rs
akirillo: Similarly to below, this value is only used in the `deploy_test_contracts` script, which creates the arguments for the `deploy_erc20` script manually. I'll document the `deploy_test_contracts` method to specify that it's intended for usage with the integration tests in this repo.

PR #204, opened 2024-01-18
scripts, integration: use system srs

akirillo:
This PR makes the necessary changes in the `scripts` & `integration` crates (as well as any testing utilities) to replace usage of any generated testing SRS with the system SRS defined in the relayer. This included removing the script to generate a testing SRS, and removing the SRS file path from any CLI args / interfaces it was previously a part of.

Additionally, I generated the production verification keys using the system SRS.

**Testing:**
All unit & integration tests pass.


contracts-utils/src/conversion.rs
```diff
@@ -13,6 +13,7 @@ use contracts_common::types::{
 use mpc_plonk::proof_system::structs::VerifyingKey;
 use mpc_relation::proof_linking::GroupLayout;
 
+// TODO: Replace w/ `to_contract_link_proof` defined in relayer when it's merged
```
akirillo: calling this out: I should implement this & merge only after https://github.com/renegade-fi/renegade/pull/340 is merged

joeykraut: merged!

scripts/src/commands.rs
```diff
@@ -398,44 +398,28 @@ pub async fn upgrade(
...
-
 /// Generates verification keys for the protocol circuits
 pub fn gen_vkeys(args: GenVkeysArgs) -> Result<(), ScriptError> {
-    let srs = parse_srs_from_file(&args.srs_path)?;
-
     let (valid_wallet_create_vkey, valid_wallet_update_vkey, match_vkeys, match_linking_vkeys) =
         if args.test {
             (
-                gen_circuit_vkey::<DummyValidWalletCreate>(&srs)
+                to_contract_vkey((*DummyValidWalletCreate::verifying_key()).clone())
```
joeykraut, line 406: what's the reason here for using `DummyValidWalletCreate` instead of the actual `ValidWalletCreate`? I know they use the same statement but it feels more direct to use the relayer type exactly. Plus generating the vkey is fast regardless of whether the constraints used are real or not

joeykraut (approved):
great!

scripts/src/commands.rs
akirillo: in the integration tests, we use the dummy circuits, and this branch of the script generates the verification keys for them. if we use the vkeys for the real circuits, the integration tests will fail

PR #206, opened 2024-01-22
contracts-stylus: darkpool: disable storage cache

akirillo:
This PR disables the Stylus storage cache for the darkpool contract. After the pubkey serialization PR, it seems the darkpool contract itself inflated just over the binary size limit. Interestingly, changing its compilation flags even to the max-compression flags used by the darkpool testing contract did not help (I think this is to do with how the `#[external]` proc macro is expanded). The simplest approach to getting the production darkpool contract within the binary size limit is to disable the storage cache. Given the storage access patterns in the darkpool, we don't get much performance out of it, anyway. This gave us enough breathing room to re-enable the max-perf compilation flags for it.

### DIFF
--- a/contracts-stylus/Cargo.toml
+++ b/contracts-stylus/Cargo.toml
@@ -16,7 +16,7 @@ alloy-sol-types = { workspace = true }
 serde = { workspace = true }
 
 [features]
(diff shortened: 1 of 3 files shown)


joeykraut (approved):
do we know the performance impact of this?

PR #207, opened 2024-02-13
contracts-stylus: darkpool: use `permit2` for deposits

akirillo:
This PR changes the Darkpool contract implementation to pull in deposits through the `permit2` contract, expecting a calldata-serialized Permit2 message and signature to be passed alongside the other arguments to `update_wallet`.

We use the [`permitTransferFrom` method](https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#single-permittransferfrom) on the Permit2 contract. A good overview of the Permit2 pattern can be found [here](https://github.com/dragonfly-xyz/useful-solidity-patterns/tree/main/patterns/permit2).

Testing is deferred to the next PR for brevity.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -2,6 +2,7 @@
 //! verifying the various proofs of the Renegade protocol, and handling deposits / withdrawals.
 
 use alloc::{vec, vec::Vec};
+use alloy_sol_types::SolType;
 use contracts_common::types::{
     ExternalTransfer, MatchPayload, PublicSigningKey, ScalarField, ValidMatchSettleStatement,
     ValidWalletCreateStatement, ValidWalletUpdateStatement,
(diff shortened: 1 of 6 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -601,18 +614,47 @@ impl DarkpoolContract {
...
+            call_helper::<permitTransferFromCall>(
+                storage,
+                permit2_address /* address */,
+                (
+                    permit,
+                    signature_transfer_details,
+                    transfer.account_addr, /* owner */
+                    signature,
+                ),
+            )?;
```
akirillo, line 656: we effectively incur a serialization round-trip here on the `permit` and `signature` - this is because creating custom types that are ABI-compatible is quite cumbersome still in Stylus

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -601,18 +614,47 @@ impl DarkpoolContract {
...
-            (darkpool_address, transfer.account_addr)
+
+        if transfer.is_withdrawal {
+            // In the case of a withdrawal, we make a simple `transfer` call
+            // from the darkpool to the user.
+            call_helper::<transferCall>(
+                storage,
+                transfer.mint /* address */,
+                (
+                    transfer.account_addr /* from */,
```
joeykraut, line 628: nit:
```suggestion
                    transfer.account_addr /* to */,
```

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: fixed in #209

PR #209, opened 2024-02-15
integration: external transfer test w/ permit2 passing

akirillo:
This PR updates the external transfer test to employ the Permit2 pattern by appropriately encoding and signing a `PermitTransferFrom` struct. We do this both in the case of deposit & withdrawal for simplicity, but of course it is only used contract-side for deposits.

Implementing & running this test led to another change, namely that we now use `serde` for serialization of the relevant Permit2 data that must be passed to the contract, which is just the permit nonce, deadline, and signature. Everything else can be gathered from the external transfer struct. This technically limits the Permit2 functionality, one could approve more funds than necessary for transfer, but now we implicitly constrain that the permitted funds must be exactly equal to what is in the external transfer struct.

The reason for this change is binary size - using `serde` in this way led to a smaller binary for the darkpool / darkpool testing contract than the ABI serialization round trip we were invoking before, and smaller even than passing the nonce, deadline, and signature as separate method arguments to `update_wallet` (which clutters the interface, anyway).

All integration tests pass.

### DIFF
--- a/contracts-common/src/solidity.rs
+++ b/contracts-common/src/solidity.rs
@@ -1,0 +1,72 @@
+//! Solidity type definitions used throughout the project
+
+#![allow(missing_docs)]
+
+use alloc::vec::Vec;
+use alloy_sol_types::sol;
+
+// Types & methods from the Permit2 `ISignatureTransfer` interface, taken from https://github.com/Uniswap/permit2/blob/main/src/interfaces/ISignatureTransfer.sol
+sol! {
+    /// The token and amount details for a transfer signed in the permit transfer signature
+    struct TokenPermissions {
+        /// ERC20 token address
+        address token;
+        /// the maximum amount that can be spent
+        uint256 amount;
+    }
+
+    /// The signed permit message for a single token transfer
+    ///
+    /// NOTE: This differs from the `PermitTransferFrom` struct in the `ISignatureTransfer` interface
+    /// in that it includes the `spender` field. This field is signed and thus must be included in the
+    /// EIP-712 hash, but is not included in the Solidity definition of the  `PermitTransferFrom` struct
+    /// (as this field is injected by the Permit2 contract).
+    struct PermitTransferFrom {
+        /// The token permissions for the transfer
+        TokenPermissions permitted;
+        /// The address to which the transfer is made
+        address spender;
+        /// a unique value for every token owner's signature to prevent signature replays
+        uint256 nonce;
+        /// deadline on the permit signature
+        uint256 deadline;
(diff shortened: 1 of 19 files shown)


contracts-common/src/solidity.rs
akirillo, line 1: this was copied over from `contracts-stylus/src/utils/solidity.rs`, but the `PermitPayload` struct was removed.

scripts/src/utils.rs
```diff
@@ -40,12 +40,15 @@ use crate::{
     solidity::initializeCall,
 };
 
+/// An Ethers provider that uses a `LocalWallet` to generate signatures
+pub type LocalWalletProvider<M> = SignerMiddleware<M, LocalWallet>;
```
akirillo, line 44: Creating this type is necessary for us to be able to access the underlying private key for signing the permit data

integration/src/utils.rs
```diff
@@ -222,3 +250,84 @@ pub(crate) fn insert_shares_and_get_root(
...
+
+/// This is a re-implementation of `eip712_signing_hash` (https://github.com/alloy-rs/core/blob/v0.3.1/crates/sol-types/src/types/struct.rs#L117)
+/// which correctly encodes the data for the nested `TokenPermissions` struct.
+///
+/// We do so by mirroring the functionality implemented in the `sol!` macro (https://github.com/alloy-rs/core/blob/v0.3.1/crates/sol-macro/src/expand/struct.rs#L56)
+/// but avoiding the (unintended) extra hash of the `TokenPermissions` struct's EIP-712 struct hash.
+///
+/// This is fixed here: https://github.com/alloy-rs/core/pull/258
+/// But the version of `alloy` used by `stylus-sdk` is not updated to include this fix.
+fn permit_signing_hash(permit: &PermitTransferFrom, domain: &Eip712Domain) -> B256 {
```
akirillo, line 315: this is IMO the best way to avoid the whole headache described above, patching the alloy deps doesn't work since `stylus-sdk` specifies its dep all the way down to a patch version. If/when `stylus-sdk` bumps their alloy deps past `v0.4.0` we can remove this code

contracts-common/src/solidity.rs
akirillo, line 1: this was copied over from `contracts-stylus/src/utils/solidity.rs`, but the `PermitPayload` struct was removed.

integration/src/utils.rs
joeykraut: oh wow haha

joeykraut (approved):

PR #210, opened 2024-02-15
contracts-stylus: darkpool: sign external transfer

akirillo:
This PR updates the darkpool to check a signature over the external transfer struct, and updates the tests to generate this signature.

Due to binary size constraints being hit, I was unable to deploy the darkpool / run the tests. Mitigations for binary size are coming in the next PR, after which I will test this functionality.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -31,7 +31,10 @@ use crate::{
             VERIFICATION_FAILED_ERROR_MESSAGE, ZERO_ADDRESS_ERROR_MESSAGE, ZERO_FEE_ERROR_MESSAGE,
         },
         helpers::{
-            assert_valid_signature, call_helper, delegate_call_helper, deserialize_from_calldata, pk_to_u256s, postcard_serialize, scalar_to_u256, serialize_match_statements_for_verification, serialize_statement_for_verification, static_call_helper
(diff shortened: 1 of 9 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -626,9 +629,18 @@ impl DarkpoolContract {
...
-        permit_payload: Bytes,
+        transfer_aux_data_bytes: Bytes,
     ) -> Result<(), Vec<u8>> {
+        let transfer_aux_data: TransferAuxData =
+            deserialize_from_calldata(&transfer_aux_data_bytes)?;
+        assert_valid_signature(
+            old_pk_root,
+            &postcard_serialize(&transfer)?,
+            &transfer_aux_data.transfer_signature,
+        )?;
```
joeykraut, line 642: why do we need to do this on the deposit path? This would require an extra signature on the FE from metamask

joeykraut (approved):

contracts-stylus/src/contracts/darkpool.rs
akirillo: ah fair point, i was doing this for implementation simplicity but didn't consider the UX. will remove

akirillo: I'll make this change in #211 to avoid the change being swallowed during merge

PR #211, opened 2024-02-15
contracts-stylus: darkpool, transfer_executor: add transfer executor contract

akirillo:
This PR introduces the `TransferExecutorContract`, which stores the Permit2 contract address and is responsible for validating & executing external transfers. With this, the Darkpool contract now deploys at 22.1kb (down from 24.7), and the transfer executor itself deploys at 10.6kb.

Changes to the `scripts` crate to deploy the `TransferExecutorContract` will come in a follow-up PR.

### DIFF
--- a/contracts-stylus/src/contracts/transfer_executor.rs
+++ b/contracts-stylus/src/contracts/transfer_executor.rs
@@ -1,0 +1,126 @@
+//! The transfer executor contract, responsible for executing external transfers to/from the darkpool.
+
+use crate::utils::{
+    constants::{MERKLE_STORAGE_GAP_SIZE, MISSING_TRANSFER_AUX_DATA_ERROR_MESSAGE},
+    helpers::{assert_valid_signature, call_helper, deserialize_from_calldata, postcard_serialize},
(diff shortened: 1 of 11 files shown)


contracts-stylus/src/contracts/transfer_executor.rs
```diff
@@ -0,0 +1,126 @@
...
+    ) -> Result<(), Vec<u8>> {
+        let old_pk_root: PublicSigningKey = deserialize_from_calldata(&old_pk_root)?;
+        let transfer: ExternalTransfer = deserialize_from_calldata(&transfer)?;
+        let transfer_aux_data: TransferAuxData = deserialize_from_calldata(&transfer_aux_data)?;
+
+        assert_valid_signature(
+            &old_pk_root,
+            &postcard_serialize(&transfer)?,
+            &transfer_aux_data.transfer_signature,
+        )?;
```
joeykraut: same q as the last pr, why do we do this for deposits?

joeykraut (approved):

PR #215, opened 2024-02-22
contracts-common, contracts-stylus: ingest fee design changes to existing protocol

akirillo:
This PR makes the relevant changes from the fee design to the existing protocol (i.e., without introducing the new statements / circuits for fees).

Namely:
- Changed types:
    - `ValidWalletUpdateStatement`
        - [x]  No longer has `timestamp`
    - `ValidMatchSettleStatement`
        - [x]  Now has `protocol_fee`
- Changed functionality:
    - `process_match_settle`
        - [x]  Ensure that same indices are used between the parties’ `ValidCommitmentsStatement`s & the `ValidMatchSettleStatement`
        - [x]  Ensure that the proper `protocol_fee` is used

The testing utilities were also updated to correctly generate statement types, and new tests were added to ensure that the new out-of-circuit checks in `process_match_settle` are working properly. All integration and unit tests pass.

Note, that these tests pass when using a local patch of the relayer crate that makes the necessary/minimal/incomplete changes to ensure that conversion logic & verification key generation work correctly.


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -413,14 +410,27 @@ impl DarkpoolContract {
...
-            match_linking_proofs,
-        )?);
+        if_verifying!({
+            assert_result!(
+                party_0_match_payload.valid_commitments_statement.indices
+                    == valid_match_settle_statement.party0_indices
+                    && party_1_match_payload.valid_commitments_statement.indices
+                        == valid_match_settle_statement.party1_indices,
+                INVALID_ORDER_SETTLEMENT_INDICES_ERROR_MESSAGE
+            )?;
```
joeykraut: nit: this might read nicer if we broke up the compound statement into individual ones:
```rust
let party0_same_indices = party_0_match_payload.valid_commitments_statement.indices == valid_match_settle_statement.party0_indices;
let party1_same_indices = party_1_match_payload.valid_commitments_statement.indices == valid_match_settle_statement.party1_indices;

assert_result!(party0_same_indices && party1_same_indices, <error message>);
```

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -413,14 +410,27 @@ impl DarkpoolContract {
...
+
+            let protocol_fee = u256_to_scalar(storage.borrow_mut().protocol_fee.get())?;
+            assert_result!(valid_match_settle_statement.protocol_fee == protocol_fee, INVALID_PROTOCOL_FEE_ERROR_MESSAGE)?;
```
joeykraut: nit: will you document that the protocol fee we draw from storage is in fixed point representation?

joeykraut (approved):

PR #216, opened 2024-02-22
contracts-stylus: darkpool: VALID RELAYER FEE SETTLEMENT contract method

akirillo:
This PR adds the `settle_online_relayer_fee` method, representing the on-chain portion of the online settlement of a relayer fee via the `VALID RELAYER FEE SETTLEMENT` circuit.

This makes reasonable assumptions about the structure / ordering of fields in the statement type for the circuit, as can be seen in the definition of the `ValidRelayerFeeSettlementStatement` type and its serialization to scalars for verification.

Notably, these changes push the darkpool contract beyond 24kb (in large part due to new serialization code for the statement type), so it is not yet deployable, and as such I have deferred writing tests for this circuit (including scaffolding such as generating verification keys) to after I have implemented all the other fees changes. This way I'll first have a complete picture of the binary profile.

### DIFF
--- a/contracts-common/src/custom_serde.rs
+++ b/contracts-common/src/custom_serde.rs
@@ -11,7 +11,10 @@ use ark_serialize::Flags;
 use crate::{
     constants::{NUM_BYTES_ADDRESS, NUM_BYTES_FELT, NUM_BYTES_U64, NUM_SCALARS_PK, NUM_U64S_FELT},
     types::{
-        ExternalTransfer, G1Affine, G1BaseField, G2Affine, G2BaseField, MontFp256, OrderSettlementIndices, PublicInputs, PublicSigningKey, ScalarField, ValidCommitmentsStatement, ValidMatchSettleStatement, ValidReblindStatement, ValidRelayerFeeSettlement, ValidWalletCreateStatement, ValidWalletUpdateStatement
+        ExternalTransfer, G1Affine, G1BaseField, G2Affine, G2BaseField, MontFp256,
+        OrderSettlementIndices, PublicInputs, PublicSigningKey, ScalarField,
+        ValidCommitmentsStatement, ValidMatchSettleStatement, ValidReblindStatement,
+        ValidRelayerFeeSettlementStatement, ValidWalletCreateStatement, ValidWalletUpdateStatement,
     },
 };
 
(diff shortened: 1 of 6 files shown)


contracts-common/src/custom_serde.rs
```diff
@@ -273,6 +276,21 @@ impl ScalarSerializable for ValidMatchSettleStatement {
...
+    fn serialize_to_scalars(&self) -> Result<Vec<ScalarField>, SerdeError> {
+        let mut scalars: Vec<ScalarField> = vec![
+            self.merkle_root,
+            self.sender_nullifier,
+            self.recipient_nullifier,
+            self.sender_wallet_commitment,
+            self.recipient_wallet_commitment,
+        ];
+        scalars.extend(&self.sender_updated_public_shares);
+        scalars.extend(&self.recipient_updated_public_shares);
```
joeykraut, line 289: we're missing `recipient_pk_root` here

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -717,39 +786,43 @@ impl DarkpoolContract {
         assert_result!(result, VERIFICATION_FAILED_ERROR_MESSAGE)
     }
 
-    /// Handles the post-match-settle logic for a single party
-    pub fn process_party<S: TopLevelStorage + BorrowMut<Self>>(
+    /// Nullifies the old wallet and commits to the new wallet,
+    /// verifying a signature over the commitment to the new wallet
+    /// if necessary.
+    pub fn rotate_wallet<S: TopLevelStorage + BorrowMut<Self>>(
```
joeykraut: might be nicer at callsites to split this out into `rotate_wallet` and `rotate_wallet_with_signature` wherein the ladder calls for the former after verifying the signed commitment.

joeykraut (approved):

contracts-common/src/custom_serde.rs
akirillo: great catch ty

contracts-stylus/src/contracts/darkpool.rs
akirillo: yeah i like it, shouldn't affect binary size either idt

PR #217, opened 2024-02-23
contracts-stylus: darkpool: VALID OFFLINE FEE SETTLEMENT contract method

akirillo:
This PR adds the `settle_offline_fee` method to the darkpool, encapsulating the contract logic involved in settling an offline fee into a note via the `VALID OFFLINE FEE SETTLEMENT` circuit.

I make reasonable assumptions on the structure and ordering of the `ValidOfflineFeeSettlementStatement` statement fields.

This PR adds the protocol public encryption key as a settable storage variable on the contract. We assert that the pubkey in the statement matches this, as opposed to injecting it into the statement. The reason for this is that in the case that the protocol pubkey is incorrect, the contract can return a verbose error indicating this, as opposed to an opaque failure to verify.

Finally, this required an extra method on the Merkle contract to allow for inserting a note commitment directly to the tree.

As before, testing is deferred until all fees changes are made and the contracts are deployable.

### DIFF
--- a/contracts-stylus/src/contracts/merkle.rs
+++ b/contracts-stylus/src/contracts/merkle.rs
@@ -107,7 +107,12 @@ where
 
         let shares_commitment = self.compute_shares_commitment(shares)?;
 
-        self.insert_helper(shares_commitment, P::HEIGHT as u8, insert_index, true)?;
+        self.insert_helper(
+            shares_commitment,
+            P::HEIGHT as u8,
+            insert_index,
+            true, /* subtree_filled */
+        )?;
 
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


contracts-stylus/src/contracts/merkle.rs
```diff
@@ -155,6 +155,24 @@ where
...
+        assert_result!(
+            insert_index < 2_u128.pow(P::HEIGHT as u32),
+            TREE_FULL_ERROR_MESSAGE
+        )?;
+
+        self.insert_helper(
+            u256_to_scalar(note_commitment)?,
+            P::HEIGHT as u8,
+            insert_index,
+            true,
```
joeykraut: nit: Can we add an inline as to what this value means?

joeykraut (approved):

PR #233, opened 2024-03-05
contracts-stylus: verification disablement guard

akirillo:
This PR adds a check to the `if_verifying!` macro which, in the case of verification being disabled, reverts unless the current chain ID is that of our designated devnet chain ID, which I've chosen to be `473474` (first six digits of keccak256("renegade").

This PR also brings the committed verification keys up to date with `renegade/main`, and includes a tweak to the deploy scripts ensuring that subsequent deployments don't try to re-optimize an old optimized WASM artifact (I ran into this issue during some local testing).

I tested that the contracts appropriately revert / allow transactions on a verification-disabled build by running sequencers w/ the old/new chain IDs.


contracts-common/src/constants.rs
```diff
@@ -6,6 +6,10 @@ use ark_ff::{BigInt, Fp};
 
 use crate::types::ScalarField;
 
+/// Dedicated chain ID for Renegade devnets, for which we allow
+/// verification to be disabled
+pub const DEVNET_CHAINID: u64 = 473474;
```
joeykraut, line 12: nit: can we document how this chain id was derived here? The context in the PR description is good

joeykraut (approved):

PR #234, opened 2024-03-30
contracts-stylus: transfer_executor: include `pk_root` as witness data for Permit2

akirillo:
This PR addresses the vulnerability disclosed in our audit concerning the ability to "frontrun" a deposit by using a properly signed `permitTransferFrom` payload but generating a proof wherein the deposit is directed to the attacker's wallet within the Renegade protocol.

We mitigate this by including the intended recipient wallet's `pk_root` as additional witness data to be signed, validated by the Permit2 contract by leveraging its [`permitWitnessTransferFrom`](https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#single-permitwitnesstransferfrom) method.

I have added an integration test covering this specific case. All `test_external_transfer*` tests pass.

### DIFF
--- a/contracts-common/src/constants.rs
+++ b/contracts-common/src/constants.rs
@@ -113,7 +113,7 @@ pub const INVALID_INPUTS_ERROR_MESSAGE: &[u8] = b"invalid inputs";
 /// operation fails
 pub const ARITHMETIC_BACKEND_ERROR_MESSAGE: &[u8] = b"arithmetic backend error";
 
-/// The witness type string used for deposits via `permitWitnessTransferFrom`.
+/// The EIP-712 type string used for deposit witness data via `permitWitnessTransferFrom`.
 ///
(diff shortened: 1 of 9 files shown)


contracts-common/src/constants.rs
```diff
@@ -112,3 +112,8 @@ pub const INVALID_INPUTS_ERROR_MESSAGE: &[u8] = b"invalid inputs";
 /// The revert message when an EC arithmetic backend
 /// operation fails
 pub const ARITHMETIC_BACKEND_ERROR_MESSAGE: &[u8] = b"arithmetic backend error";
+
+/// The witness type string used for deposits via `permitWitnessTransferFrom`.
```
joeykraut: nit: can we note that this is an EIP-712 type string?

joeykraut (approved):

PR #240, opened 2024-04-25
contracts-stylus: merkle: Add leaf insertion event

joeykraut:
### Purpose
This PR adds a leaf node insertion event. Now that we have changed the `MerkleOpeningNode` event to be the siblings in the path, we need an index of the insertions themselves to find commitments in the tree. The `MerkleInsertion` event fills this need.

### Testing
- Deployed to a local sequencer, able to reconstruct Merkle paths with updated client logic

### DIFF
--- a/contracts-stylus/src/utils/solidity.rs
+++ b/contracts-stylus/src/utils/solidity.rs
@@ -54,6 +54,7 @@ sol! {
 
     // Merkle events; we emit the opening path of the inserted node
     event MerkleOpeningNode(uint8 indexed height, uint128 indexed index, uint256 indexed new_value);
+    event MerkleInsertion(uint128 indexed index, uint256 indexed value);
 
     // Darkpool user interaction events
     event NullifierSpent(uint256 indexed nullifier);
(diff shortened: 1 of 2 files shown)


contracts-stylus/src/utils/solidity.rs
```diff
@@ -54,6 +54,7 @@ sol! {
 
     // Merkle events; we emit the opening path of the inserted node
     event MerkleOpeningNode(uint8 indexed height, uint128 indexed index, uint256 indexed new_value);
```
akirillo, line 56: pretty sure we no longer need the `index` field in the `MerkleOpeningNode` event, right? if so we should remove it, this will reduce binary a bit by reducing the encoding/decoding logic the `sol!` macro expands to for this event

akirillo (approved):

contracts-stylus/src/utils/solidity.rs
joeykraut: Disagree, when we want to add back in the updating Merkle paths logic we will need this. It's also not something we can really add after the fact so it should be in day 1.

PR #246, opened 2024-08-14
contracts-stylus: darkpool-core: Track public blinders, disallow reuse

joeykraut:
### Purpose
This PR adds a `public_bilnder_set` to the `DarkpoolCoreContract` in a fashion that mirrors the nullifier set implementation. That is, we expose a getter on the `DarkpoolContract` and check that a public blinder is unused in the `check_wallet_rotation` method.

Testing will come in a follow-up, we need to fix a few issues in the integration testing stack first.

### DIFF
--- a/contracts-stylus/src/contracts/darkpool_core.rs
+++ b/contracts-stylus/src/contracts/darkpool_core.rs
@@ -12,14 +12,14 @@ use crate::{
             CALL_RETDATA_DECODING_ERROR_MESSAGE, INVALID_ARR_LEN_ERROR_MESSAGE,
             INVALID_ORDER_SETTLEMENT_INDICES_ERROR_MESSAGE, INVALID_PROTOCOL_FEE_ERROR_MESSAGE,
             INVALID_PROTOCOL_PUBKEY_ERROR_MESSAGE, MERKLE_STORAGE_GAP_SIZE,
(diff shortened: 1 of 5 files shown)


contracts-stylus/src/contracts/darkpool_core.rs
```diff
@@ -705,6 +731,8 @@ impl DarkpoolCoreContract {
         merkle_root: ScalarField,
         new_wallet_public_shares: &[ScalarField],
     ) -> Result<(), Vec<u8>> {
+        let public_blinder = get_public_blinder_from_shares(new_wallet_public_shares);
```
akirillo, line 735: tiny nit: can we update the doc comment above to mention that the new wallet's public blinder is checked

contracts-stylus/src/contracts/darkpool_core.rs
```diff
@@ -753,7 +781,8 @@ impl DarkpoolCoreContract {
     pub fn log_wallet_update(public_wallet_shares: &[ScalarField]) {
         // We assume the wallet blinder is the last scalar serialized into the wallet shares.
         // Unwrapping here is safe because we know the wallet shares are non-empty.
-        let wallet_blinder_share = scalar_to_u256(*public_wallet_shares.last().unwrap());
+        let wallet_blinder_share =
```
akirillo, line 783: nit: can we remove the comment above, this assumption is now embodied by `get_public_blinder_from_shares`

akirillo (approved):
lgtm, couple tiny doc nits

PR #248, opened 2024-08-17
contracts-stylus: utils: ABI encode staticcall calldata

joeykraut:
### Purpose
This PR fixes a bug in how static calls are handled that surfaced as a result of `alloy` upgrades. As well, now that the contracts are in a working state with the new `alloy` version, I added an integration test for public blinder uniqueness, using the `new_wallet` selector as a representative.

### Testing
- Tested the integration test, checked that it failed if I didn't re-use the blinder
- Deployed all contracts and tested against a local relayer


akirillo (approved):
slick my g

PR #256, opened 2024-10-09
contracts-stylus: vkeys: Add atomic match settle vkey

joeykraut:
### Purpose
This PR adds the verification bundle for `VALID MATCH SETTLE ATOMIC` to the `vkeys` contract, along with a method to fetch it. This bundle includes:
- Verification keys for `VALID REBLIND`, `VALID COMMITMENTS`, and `VALID MATCH SETTLE ATOMIC`
- Proof-link verification keys for `VALID REBLIND <-> VALID COMMITMENTS` and `VALID COMMITMENTS <-> VALID MATCH SETTLE ATOMIC`

### Testing
- Built and deployed vkeys contract


contracts-utils/src/proof_system/mod.rs
```diff
@@ -97,6 +117,21 @@ pub fn gen_match_linking_vkeys<C: SingleProverCircuit>(
     })
 }
 
+/// Generate the verification keys for the circuits involved in settling a matched trade
```
akirillo:
```suggestion
/// Generate the linking verification keys for the circuits involved in settling a matched atomic trade
```
:P

akirillo (approved):

PR #257, opened 2024-10-09
contracts-stylus: contract: darkpool: Add atomic match settle method

joeykraut:
### Purpose
This PR adds a method to verify an atomic match settlement. This follows mostly the same flow as the standard match settlement, except that only _one_ party has a `VALID COMMITMENTS` and `VALID REBLIND` bundle.

Note that, in this state, the `darkpool-core` contract is too large, so a follow-up will optimize this below the binary size limit.

### Testing
- Deferred until complete

### DIFF
--- a/contracts-common/src/types.rs
+++ b/contracts-common/src/types.rs
@@ -255,6 +255,41 @@ pub struct TransferAuxData {
     pub transfer_signature: Option<Vec<u8>>,
 }
 
+/// A fee take from a match
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+pub struct FeeTake {
+    /// The fee the relayer takes
+    #[serde_as(as = "U256Def")]
+    pub relayer_fee: U256,
+    /// The fee the protocol takes
+    #[serde_as(as = "U256Def")]
+    pub protocol_fee: U256,
+}
+
+/// The result of an atomic match
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+pub struct ExternalMatchResult {
+    /// The mint (erc20 address) of the quote token
+    #[serde_as(as = "AddressDef")]
+    pub quote_mint: Address,
+    /// The mint (erc20 address) of the base token
+    #[serde_as(as = "AddressDef")]
+    pub base_mint: Address,
+    /// The amount of the quote token
+    #[serde_as(as = "U256Def")]
+    pub quote_amount: U256,
+    /// The amount of the base token
+    #[serde_as(as = "U256Def")]
+    pub base_amount: U256,
+    /// The direction of the trade
+    ///
+    /// `false` (0) corresponds to the internal party buying the base
+    /// `true` (1) corresponds to the internal party selling the base
+    pub direction: bool,
+}
+
 /// Represents the affine coordinates of a secp256k1 ECDSA public key.
 /// Since the secp256k1 base field order is larger than that of Bn254's scalar field,
 /// it takes 2 Bn254 scalar field elements to represent each coordinate.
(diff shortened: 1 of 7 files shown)


contracts-common/src/types.rs
```diff
@@ -533,6 +590,17 @@ pub struct MatchLinkingWirePolyComms {
     pub valid_match_settle: G1Affine,
 }
 
+/// The public inputs for the `MatchAtomicProofs`
+#[derive(Serialize, Deserialize)]
+pub struct AtomicMatchSettlePublicInputs {
```
akirillo, line 595: will we also use an analogous `AtomicMatchSettleLinkingWirePolyComms`, or will we use the existing `MatchLinkingWirePolyComms` and set dummy values for `valid_reblind_1` and `valid_commitments_1`? I'd prefer the former just to have clearer types, it should have a negligible impact on binary size

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -443,6 +443,37 @@ impl DarkpoolContract {
...
     }
 
+    /// Processes an atomic match settlement between two parties; one internal and one external
+    ///
+    /// An internal party is one with state committed into the darkpool, while an external party provides liquidity to the pool
+    /// during the transaction in which this method is called
+    ///
+    /// The `match_proofs` argument is the serialization of the [`contracts_common::types::ExternalMatchProofs`]
+    /// struct, and the `match_linking_proofs` argument is the serialization of the
+    /// [`contracts_common::types::ExternalMatchLinkingProofs`] struct
```
akirillo, line 453: the referenced `ExternalMatchProofs` and `ExternalMatchLinkingProofs` types are not present in this PR, I'm assuming you have them implemented in a follow-up?

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -443,6 +443,37 @@ impl DarkpoolContract {
...
+    pub fn process_atomic_match_settle<S: TopLevelStorage + BorrowMut<Self>>(
+        storage: &mut S,
+        internal_party_match_payload: Bytes,
+        valid_match_settle_statement: Bytes,
```
akirillo: nit: name `valid_match_settle_atomic_statement`?

contracts-stylus/src/contracts/darkpool_core.rs
```diff
@@ -282,6 +285,71 @@ impl DarkpoolCoreContract {
...
+            internal_party_match_payload
+                .valid_reblind_statement
+                .merkle_root,
+            internal_party_match_payload
+                .valid_reblind_statement
+                .reblinded_private_shares_commitment,
+            &valid_match_settle_atomic_statement.internal_party_modified_shares,
+        )?;
+
+        Ok(())
```
akirillo, line 350: totally understandable if this is left for a subsequent PR, but after verifying we'll need to execute the associated token transfers, right?

akirillo (approved):
just a couple qs about missing types which i assume are in a follow up pr, lgtm

contracts-stylus/src/contracts/darkpool.rs
joeykraut: yeah the split between which types show up here versus in the next prs got messed up. they will show up, apologies for the confusion

contracts-stylus/src/contracts/darkpool_core.rs
joeykraut: yes, exactly

contracts-common/src/types.rs
joeykraut: we'll use the former (new type) see the comment below, the spread of types across prs is a bit poorly managed

PR #258, opened 2024-10-10
contracts-core: verifier: Add match settle atomic verification method

joeykraut:
### Purpose
This PR adds a verifier method for verifying the atomic match settlement bundle. This closely follows the standard match path, but requires fewer proof-linking groups.

### Testing
Deferred to a follow up

### DIFF
--- a/contracts-core/src/verifier/mod.rs
+++ b/contracts-core/src/verifier/mod.rs
@@ -9,10 +9,12 @@ use alloc::{vec, vec::Vec};
 use ark_ff::{batch_inversion, FftField, Field, One, Zero};
 use contracts_common::{
     backends::{G1ArithmeticBackend, HashBackend},
-    constants::{NUM_MATCH_LINKING_PROOFS, NUM_WIRE_TYPES},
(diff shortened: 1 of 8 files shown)


contracts-core/src/verifier/mod.rs
```diff
@@ -315,6 +393,50 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
         })
     }
 
+    /// Computes the elements used in the final KZG batch opening pairing check
+    /// for the linking proofs involved in the matching and settlement of an atomic match.
+    fn prep_atomic_match_linking_proofs_opening(
```
akirillo, line 365: if we're going to be splitting the verifier contract anyway, i'm inclined to make this method more generic, i.e. accept vectors of vkeys/proofs/commitments & loop over them, so the same method could be used here and in the classic match-settle batch verification. without a doubt the reason this is not currently the case is b/c of binary size constraints

akirillo (approved):

contracts-core/src/verifier/mod.rs
joeykraut: i like it, i added a helper that takes in a vector of the new `LinkingInstance` type and accumulates the pairing product elements into `OpeningElems`

PR #259, opened 2024-10-10
contracts-stylus: contracts: core: Split up core contract

joeykraut:
### Purpose
This PR splits up the `DarkpoolCore` contract into a `CoreSettlementContract` and `CoreWalletOpsContract` which respectively contain:
- Logic for settling trades
- Logic for updating wallets

This brings the core contract under the binary size limit

### Testing
- Deployed both contracts to testnet


contracts-stylus/src/contracts/core/core_helpers.rs
```diff
@@ -0,0 +1,355 @@
+//! Helpers for core contracts, mostly around interacting with core state
```
joeykraut, line 1: these are basically copy pasted from `darkpool_core.rs` with modifications to use the `CoreContractStorage` layout.

akirillo (approved):

PR #261, opened 2024-10-10
contracts-stylus: verifier: Split verifier into two contracts

joeykraut:
### Purpose
This PR splits the verifier contract into two contracts so that they fit under the size limit. Specifically:
- `CoreVerifierContract`: holds the core Plonk verification logic and exposes an interface to batch verify a set of proofs with optional extra arguments for proof-linking
- `SettlementVerifierContract`: sets up the batch verification bundles for match settlement proofs. This includes adding the proof linking elements and sending them as the KZG batch opening parameter to the core verifier

### Testing
- Deployed both contracts to testnet to verify sizes
- Other testing deferred to follow up


contracts-core/src/verifier/mod.rs
```diff
@@ -50,40 +49,46 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
-    /// - `PARTY 0 VALID COMMITMENTS` <-> `VALID MATCH SETTLE`
-    /// - `PARTY 1 VALID COMMITMENTS` <-> `VALID MATCH SETTLE`
-    ///
-    /// Applies batch verification as implemented in Jellyfish: https://github.com/renegade-fi/mpc-jellyfish/blob/main/plonk/src/proof_system/verifier.rs#L199
-    ///
-    /// This assumes that all the verification keys were generated using the same SRS.
-    pub fn verify_match(
-        match_vkeys: MatchVkeys,
+    /// Prepare the linking proofs opening elements for the match linking proofs
+    pub fn get_match_linking_opening_elems(
```
akirillo: IMO, this feels like an unnecessary extra layer of indirection before `prep_match_linking_proofs_opening`. I would just have that method also accept the `match_proofs` and extract the wire poly commitments there.

That way, we have a powerful generic `prep_linking_instances`, and a _single_ method per settlement type (classic / atomic) that constructs the `LinkingInstance` inputs for it.

In fact, I'd even be in favor of getting rid of the `MatchVkeys`, `MatchLinkingVkeys`, `MatchProofs`, `MatchLinkingProofs` structs for vector types, so that we can just naively zip vkeys / proofs / commitments together (at the expense of duplicating some elements in these vectors, e.g. match-settle linking wire poly comm would have to show up twice) to make the verifier more easily extensible and have less indirection.

all of this also applies to the atomic match-settle code paths as well, of course

contracts-core/src/verifier/mod.rs
```diff
@@ -50,40 +49,46 @@ impl<G: G1ArithmeticBackend, H: HashBackend> Verifier<G, H> {
...
         proof: Proof,
         public_inputs: PublicInputs,
+    ) -> Result<bool, VerifierError> {
+        Self::batch_verify(&[vkey], &[proof], &[public_inputs], None)
+    }
+
+    /// Batch-verifies a set of proofs
+    ///
+    /// Optionally, additional opening elements may be passed to facilitate proof-linking
+    pub fn batch_verify(
```
akirillo, line 59: this makes me so happy

contracts-stylus/src/contracts/verifier/verifier_settlement.rs
```diff
@@ -0,0 +1,127 @@
...
+/// The verifier with the precompile G1 backend and Stylus hasher
+type StylusVerifier = Verifier<PrecompileG1ArithmeticBackend, StylusHasher>;
+
+/// The calldata for the `verify_match` function
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+struct VerifyMatchCalldata {
+    /// The verifier address
+    #[serde_as(as = "AddressDef")]
+    verifier_address: Address,
```
akirillo: i understand having this in calldata to keep the verifier stateless - IMO this is nice mostly bc we can avoid access controls around setting state variables - but if/when Stylus contract constructors ever become a thing, i would prefer having the core verifier address be a state variable in the settlement verifier contract that gets set in the constructor (and never again)

contracts-stylus/src/contracts/verifier/verifier_settlement.rs
```diff
@@ -0,0 +1,127 @@
...
+    helpers::{deserialize_from_calldata, postcard_serialize, static_call_helper},
+};
+
+/// The verifier with the precompile G1 backend and Stylus hasher
+type StylusVerifier = Verifier<PrecompileG1ArithmeticBackend, StylusHasher>;
+
+/// The calldata for the `verify_match` function
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+struct VerifyMatchCalldata {
```
akirillo: can we move these calldata types to `contracts-common/types.rs`? that's where all such types can be discovered currently, and we'll probably need to export these types for use in integration tests (if not in the darkpool contract)

akirillo (approved):
have some opinions on indirection in these match/atomic match batch verification code paths, but lgtm

contracts-stylus/src/contracts/verifier/verifier_settlement.rs
joeykraut: will add a todo

contracts-core/src/verifier/mod.rs
joeykraut: yep, good call

PR #264, opened 2024-10-12
integration: tests: Add tests for `process_valid_match_settle_atomic`

joeykraut:
### Purpose
This PR adds a few tests for `process_valid_match_settle_atomic` following the analogous tests for `process_match_settle`. Once I add erc20 transfers, I'll test those bits as well

### Testing
- All tests pass


contracts-common/src/types.rs
```diff
@@ -765,15 +765,13 @@ pub struct VerifyAtomicMatchCalldata {
     #[serde_as(as = "AddressDef")]
     pub verifier_address: Address,
     /// The match atomic vkeys
-    pub match_atomic_vkeys: MatchAtomicVkeys,
-    /// The match atomic linking vkeys
-    pub match_atomic_linking_vkeys: MatchAtomicLinkingVkeys,
+    pub match_atomic_vkeys: Vec<u8>,
```
akirillo, line 768: are we still leaving the optimization of calldata serialization as a TODO, or are we happy w this setup?

akirillo (approved):

contracts-common/src/types.rs
joeykraut: it doesn't incur much in the way of extra gas, so I may leave it, but going to take a look in a follow up

PR #266, opened 2024-10-12
contracts-stylus: transfer-executor: Execute atomic settlement transfers

joeykraut:
### Purpose
This PR executes the transfers required to fulfill an atomic match settlement. This includes:
- One transfer for the relayer's fee on the external party's receive mint
- One transfer for the protocol's fee on the external party's receive mint
- One transfer for the external party to receive their buy mint
- One transfer for the external party to send their sell mint

### Testing
Deferred to a follow-up


contracts-common/src/types.rs
```diff
@@ -347,6 +347,50 @@ pub struct TransferAuxData {
     pub transfer_signature: Option<Vec<u8>>,
 }
 
+/// A simple erc20 transfer
+///
+/// For deposits, we directly use the erc20 contracts `transfer` function assuming that
+/// the caller has approved the darkpool contract to spend the deposit. This means that no
+/// permit2 logic is needed.
```
akirillo, line 354: this tracks, but for uncontextualized readers, can we add a callout that this generally should *only* be used to facilitate transfers for external parties in atomic matches? it is only in this case that we don't have to use permit2 as there is no risk of an erc20 transfer being deposited into the wrong wallet since the external party has no renegade wallet.

akirillo: also, i'm not terribly opinionated about this, but curious why you opted to create this struct as opposed to using the `ExternalTransfer` struct (which has the same exact fields). the fact that permit2 signatures are not expected can be documented on the associated `TransferExecutor` contract methods

contracts-stylus/src/contracts/core/core_settlement.rs
```diff
@@ -397,4 +409,69 @@ impl CoreSettlementContract {
...
+    }
+
+    /// Call the transfer executor to execute the transfers
+    fn execute_transfers<S: TopLevelStorage + BorrowMut<Self>>(
+        storage: &mut S,
+        transfers: Vec<SimpleErc20Transfer>,
+    ) -> Result<(), Vec<u8>> {
+        let transfer_executor_address = storage.borrow_mut().transfer_executor_address();
+        let calldata = postcard_serialize(&transfers)?;
+        call_helper::<executeTransferBatchCall>(
```
akirillo: we have to use `delegate_call_helper` here. using `call_helper` means we no longer use the context of the darkpool proxy contract when executing `TransferExecutorContract::execute_transfer_batch`. This means it will try to deposit to / withdraw from its _own_ contract address, instead of that of the darkpool proxy

akirillo (changes requested):
we need to switch use of a `call` to a `delegatecall`. otherwise lgtm, just have some qs

akirillo (approved):

PR #267, opened 2024-10-14
integration: Add transfer checks to atomic match settle tests

joeykraut:
### Purpose
This PR adds integration tests to the contracts to check that the external party's settlement is properly executed. That is, we check the external party's wallet balances before and after match, as well as those provided by the relayer and protocol to collect fees.

### Testing
- All integration tests pass


integration/src/main.rs
```diff
@@ -53,14 +57,44 @@ pub struct TestArgs {
...
+    ) -> Result<U256> {
+        let contract = DummyErc20Contract::new(erc20_address, self.client.clone());
+        contract
+            .balance_of(address)
+            .call()
+            .await
+            .map_err(Into::into)
+            .map(u256_to_alloy_u256)
+    }
+}
```
akirillo: nit: can we just have `TestArgs::get_erc20_balance` call `get_erc20_balance_of(self.client.address())`, and can we have `get_erc20_balance_of` be a helper in `utils.rs` that is not implemented on `TestArgs` and just takes in a `&LocalWalletHttpClient`? doesn't quite make sense to me for this method to be implemented on the testing arguments struct

integration/src/tests.rs
```diff
@@ -63,6 +67,107 @@ use crate::{
...
+    let amount = TEST_FUNDING_AMOUNT;
+    contract
+        .approve(test_args.darkpool_proxy_address, amount.into())
+        .send()
+        .await?
+        .await?;
+
+    Ok(())
+}
+
```
akirillo: nit: can we move all of this into `integration/src/utils.rs`?

integration/src/tests.rs
```diff
@@ -1281,9 +1386,13 @@ async fn test_process_match_settle__inconsistent_fee(test_args: TestArgs) -> Res
 }
 integration_test_async!(test_process_match_settle__inconsistent_fee);
 
-/// Test a successful call to `process_atomic_match_settle`
-async fn test_process_atomic_match_settle(test_args: TestArgs) -> Result<()> {
-    let contract = DarkpoolTestContract::new(test_args.darkpool_proxy_address, test_args.client);
+/// Setup an atomic match settle test
+async fn setup_atomic_match_settle_test(
```
akirillo, line 1323: partial to moving this to `utils.rs` as well, so that `tests.rs` contains only integration test code

akirillo (approved):
just some ocd nits, lgtm

integration/src/main.rs
joeykraut: can definitely do the first suggestion.

The second one I disagree with. The `TestArgs` feel useful as a generic dependency injection container, and these helpers are really only useful for injecting dependencies. I'd agree with us being stricter on our abstractions if this was not integration testing code.

PR #271, opened 2024-11-08
contracts-stylus: transfer-executor: Allow native ETH transfers

joeykraut:
### Purpose
This PR allows native ETH transfers by specifying the address `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`, which is the standard native address fill-in for swap APIs.

In the case of a native ETH deposit, the contract expects the transaction payable amount to be the amount of the transfer, and makes a call to the configured WETH contract's `deposit` method using this same `value`. 

In the case of a native ETH withdrawal, the contract will call the WETH contract's `withdrawTo` method, passing the recipient along with this call.

In order to specify the WETH contract address, we constantize the value, but leave it empty in committed code. We also add a compile-time assertion that this value is non-empty, which fails in the `main` branch. This prevents someone who is deploying with the wrong WETH address configured, as they must fill it in for the contract to compile.

### Testing
- Deployed transfer executor to check binary size
- Other testing deferred until dummy WETH contract is built

### DIFF
--- a/contracts-stylus/src/contracts/transfer_executor.rs
+++ b/contracts-stylus/src/contracts/transfer_executor.rs
@@ -1,6 +1,8 @@
 //! The transfer executor contract, responsible for executing external transfers to/from the darkpool
 //! (it is intended to be delegate-called by the darkpool)
 
+use core::str::FromStr;
+
 use crate::{
     if_verifying,
     utils::{
@@ -11,7 +13,10 @@ use crate::{
         helpers::{
             assert_valid_signature, call_helper, deserialize_from_calldata, postcard_serialize,
         },
-        solidity::{transferCall, transferFromCall, ExternalTransfer as ExternalTransferEvent},
+        solidity::{
+            depositCall, transferCall, transferFromCall, withdrawToCall,
+            ExternalTransfer as ExternalTransferEvent,
+        },
     },
 };
 use alloc::{string::ToString, vec::Vec};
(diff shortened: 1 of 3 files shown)


contracts-stylus/src/contracts/transfer_executor.rs
```diff
@@ -28,15 +33,55 @@ use contracts_common::{
...
+/// We set this to zero to avoid the contract being compiled unless this value is changed. This
+/// prevents a contract from being deployed without the correct WETH address having been set
+/// manually by the deployer.
+///
+/// For convenience, the addresses of WETH on Arbitrum chains are below:
+/// - Sepolia: 0x980B62Da83eFf3D4576C647993b0c1D7faf17c73 // TODO: replace with dummy WETH address
+/// - Arbitrum One: 0x82aF49447D8a07e3bd95BD0d56f35241523fBabb
+///
+/// WARNING: A concrete WETH address should _not_ be committed to main, but should be manually set each time by the deployer
+const WETH_ADDRESS: &str = "";
```
akirillo: IMO we should have this be read from an environment variable at compile-time using `env!`, this will make it easier to deploy this contract in a script

contracts-stylus/src/contracts/transfer_executor.rs
```diff
@@ -28,15 +33,55 @@ use contracts_common::{
...
-    contract, evm,
+    call::Call,
+    contract, evm, msg,
     prelude::*,
     storage::{StorageAddress, StorageArray, StorageU256},
 };
 
+/// A dummy address for the native asset, constant across all chains
+/// 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE
+const NATIVE_ETH_ADDRESS: &str = "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE";
```
akirillo, line 44: 🐬

contracts-stylus/src/contracts/transfer_executor.rs
```diff
@@ -28,15 +33,55 @@ use contracts_common::{
...
+fn get_weth_address() -> Address {
+    Address::from_str(WETH_ADDRESS).expect("WETH_ADDRESS must be a valid address")
+}
+
+/// A helper method to parse the native ETH address from a string
+fn is_native_eth_address(addr: Address) -> bool {
+    let native_addr =
+        Address::from_str(NATIVE_ETH_ADDRESS).expect("NATIVE_ETH_ADDRESS must be a valid address");
+    addr == native_addr
+}
```
akirillo: this is a ridiculous nit, I know, but if we start getting close to binary size limit, I would replace the `expect`s with `unwrap` as we don't really need these `expect` message string literals taking up binary size

akirillo (approved):

contracts-stylus/src/contracts/transfer_executor.rs
joeykraut: fair, zero cost to do so

PR #273, opened 2024-11-09
contracts-stylus: test-contracts: dummy-weth: Add WETH contract

joeykraut:
### Purpose
This PR adds a dummy WETH contract (which inherits the Erc20 contract) and updates the scripts to make it deployable through the CLI.

I bumped the stylus sdk version to get inheritance support, which required changing the macros that derive contract methods.

### Testing
- Deployed the WETH contract and tested its wrapper-specific methods

### DIFF
--- a/contracts-stylus/src/contracts/test_contracts/dummy_weth.rs
+++ b/contracts-stylus/src/contracts/test_contracts/dummy_weth.rs
@@ -1,0 +1,89 @@
+//! A mock ERC20 token implementation with wrapper functionality used for testing.
+//!
+//! Modeled after the WETH9 contract code:
+//!     https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code
+//!
+//! THIS IS NOT MEANT TO BE DEPLOYED AS A PRODUCTION CONTRACT.
+
+use alloc::vec::Vec;
+use alloy_sol_types::sol;
+use stylus_sdk::{
+    alloy_primitives::{Address, U256},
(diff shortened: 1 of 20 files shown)


contracts-stylus/src/lib.rs
```diff
@@ -11,7 +11,3 @@ mod contracts;
 mod utils;
 
 extern crate alloc;
-
```
joeykraut, line 14: the new SDK declares this directly

contracts-stylus/src/contracts/test_contracts/dummy_weth.rs
```diff
@@ -0,0 +1,89 @@
...
+        let amount = amount.try_into().unwrap();
+        let mut bal = self.erc20.balances.setter(sender);
+        let old_bal = bal.get();
+        if old_bal < amount {
+            return Err(ERR_WITHDRAWAL_EXCEEDS_BALANCE.to_vec());
+        }
+        bal.set(old_bal - amount);
+
+        // Transfer the ETH and log the withdrawal
+        transfer_eth(to, amount)?;
```
akirillo, line 53: we could also add a check that the contract's ETH balance is greater than `amount`, though of course the call would just revert anyway if this is not the case

akirillo (approved):

contracts-stylus/src/contracts/test_contracts/dummy_weth.rs
joeykraut: will do

PR #274, opened 2024-11-09
rust-toolchain: Bump to fix SIMD generation

joeykraut:
### Purpose
This PR bumps the rust toolchain to fix an issue wherein SIMD instructions were created in the `wasm32_unknown_unknown` target even with the `simd` feature disabled. This prevents WASM binaries from being deployed to stylus.

I also removed the darkpool contract from the list of size-optimized binaries, as it is no longer constrained by size.

### Testing
- Built and deployed all contracts


akirillo (approved):
Curious to see gas costs now that the darkpool is compiled with performance optimizations :0

PR #276, opened 2024-11-12
rustfmt: Use relayer repo's rust fmt file

joeykraut:
### Purpose
This PR adds the `rustfmt.toml` file from the relayer for repo-consistency and formats the repo with `cargo fmt`


akirillo (approved):
Approving under the assumption that the only changes here are applying the new formatting rules

PR #277, opened 2024-11-13
integration: Add integration tests for native asset external matches

joeykraut:
### Purpose
This PR adds an initial integration test to the contracts tests for native asset external transfers. This PR also fixes a couple bugs:
- All methods delegate called must be marked as `#[payable]` to accept a non-zero `value` on the transaction
- When deploying test contracts, set the WETH address environment variable as the first ERC20

### Todo
- Add further integration tests

### Testing
- Integration tests pass


contracts-stylus/src/contracts/merkle.rs
```diff
@@ -89,6 +89,7 @@ where
     }
 
     /// Returns whether or not the given root is in the root history
+    #[payable]
```
akirillo, line 92: can we add doc comments for why `#[payable]` is set on these methods in the native transfer call stack? we can just do it at the top-level `process_atomic_match_settle`, just nice to have it in the codebase for uncontextualized readers in the future

integration/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 contracts-common = { path = "../contracts-common", features = ["test-helpers"] }
 contracts-core = { path = "../contracts-core" }
+contracts-stylus = { path = "../contracts-stylus", features = ["test-helpers"] }
```
akirillo, line 9: i'm somewhat surprised this doesn't lead to compilation issues in the `integration` crate - IIRC i've tried importing from `contracts-stylus` before and had issues because the crate gets compiled to a non-WASM target. but since integration tests pass, i'm assuming this wasn't a problem?

integration/src/main.rs
```diff
@@ -32,9 +32,12 @@ mod constants;
...
 mod utils;
 
-/// The arguments provided to each integration test
+/// The context provided to each integration test
+///
+/// Allows for dependency and argument injection as well as convenient helpers
+/// for setting up tests
 #[derive(Clone)]
-pub struct TestArgs {
+pub struct TestContext {
```
akirillo, line 40: i like this naming change, nice

akirillo (approved):

contracts-stylus/src/contracts/merkle.rs
joeykraut: good call

integration/Cargo.toml
joeykraut: very little is exposed publicly via `contracts-stylus` currently, maybe at the time you had other wasm-specific exports in `contracts-stylus`?

PR #280, opened 2024-12-03
contracts-stylus: core-settlement: Accept receiver for atomic match

joeykraut:
### Purpose
This PR updates the `core-settlement` contract to allow a non-sender (`msg::sender()`) receiver in an external match.

### Testing
- [x] Integration tests pass
- [ ] Test against a testnet-deployed set of contracts

### DIFF
--- a/contracts-common/src/lib.rs
+++ b/contracts-common/src/lib.rs
@@ -3,7 +3,6 @@
 
 #![deny(missing_docs)]
 #![deny(clippy::missing_docs_in_private_items)]
-#![no_std]
 
 extern crate alloc;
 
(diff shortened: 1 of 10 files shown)


contracts-common/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 
 #![deny(missing_docs)]
 #![deny(clippy::missing_docs_in_private_items)]
-#![no_std]
```
joeykraut, line 6: no longer needed with new stylus sdk

akirillo (approved):

PR #281, opened 2025-01-06
contracts-stylus: core: Add external match protocol fee overrides slot

joeykraut:
### Purpose
This PR adds a storage map for external match protocol fee overrides. These allow the protocol fee to be set on an per-asset basis. This PR also adds setter and getter methods for these fee overrides.

Importantly, these overrides are only used in the external match path; as mints are not known to the protocol for "internal" matches.

### Testing
- Deferred to a follow-up

### DIFF
--- a/contracts-stylus/src/contracts/core/mod.rs
+++ b/contracts-stylus/src/contracts/core/mod.rs
@@ -49,4 +49,23 @@ pub trait CoreContractStorage {
 
     /// Get the address of the protocol external fee collection contract
     fn protocol_external_fee_collection_address(&self) -> Address;
+
+    /// Get the default protocol fee
+    fn protocol_fee(&self) -> U256;
+
+    /// Get the override protocol fee for a given asset
+    fn external_match_fee_override(&self, asset: Address) -> U256;
+
+    /// Get the protocol fee for a given asset on an external match
+    ///
+    /// We disallow fee overrides of _exactly_ zero, as this conflicts with the
+    /// `StorageMap` default.
+    fn external_match_protocol_fee(&self, asset: Address) -> U256 {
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/core/mod.rs
```diff
@@ -49,4 +49,23 @@ pub trait CoreContractStorage {
...
+    fn protocol_fee(&self) -> U256;
+
+    /// Get the override protocol fee for a given asset
+    fn external_match_fee_override(&self, asset: Address) -> U256;
+
+    /// Get the protocol fee for a given asset on an external match
+    ///
+    /// We disallow fee overrides of _exactly_ zero, as this conflicts with the
+    /// `StorageMap` default.
+    fn external_match_protocol_fee(&self, asset: Address) -> U256 {
```
akirillo, line 63: are we sure we want to disallow fee overrides of _exactly_ zero? I realize enabling this would require adding an extra field in the mapping entry for whether or not the override fee is set, but it could be nice to enable. even if we set the minimum possible fee instead, this still incurs an ERC20 transfer on the external party's behalf, the gas cost of which, while small, may hurt our competitiveness in winning routes. it also necessitates a fee payment task on the internal party's behalf - if no fees were accrued otherwise, this now adds a fair bit of latency to the withdrawal task.

akirillo (approved):

contracts-stylus/src/contracts/core/mod.rs
joeykraut: those optimizations are minuscule, I don't think it's worth chasing them

akirillo: tbf i think the changes necessary are minuscule as well, but i digress, its nbd

PR #283, opened 2025-01-23
contracts-stylus: contracts: gas sponsor contract initial impl

akirillo:
This PR introduces an initial implementation of the `GasSponsorContract`, which can be used to sponsor the gas costs of atomic (external) matches through the darkpool. This includes methods for sponsored & unsponsored matches, with and without a receiver specified, so that it can serve as a single entrypoint for all external match settlement.

This PR simply lays the groundwork for the implementation. I have not yet attempted to build, deploy, or test this contract. That will come in a follow-up PR.


contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -0,0 +1,313 @@
...
+        evm::log(Paused {});
+        Ok(())
+    }
+
+    /// Unpauses gas sponsorship
+    pub fn unpause(&mut self) -> Result<(), Vec<u8>> {
+        self.paused.set(false);
+        evm::log(Unpaused {});
+        Ok(())
+    }
```
joeykraut: these need to have some form of access control

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -0,0 +1,313 @@
...
+        Ok(self.paused.get())
+    }
+
+    // ----------------
+    // | KEY ROTATION |
+    // ----------------
+
+    /// Rotates the auth address to the given new address.
+    /// This must be authorized by a signature over the new address by the
+    /// previous auth private key.
```
joeykraut: i'd rather this just be only owner and we can just use one form of access control like we'll add for the pause,unpause

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -0,0 +1,313 @@
...
+            );
+        }
+
+        // Take note of the initial tx gas budget
+        let initial_gas = evm::gas_left();
+
+        // Verify the nonce & signature, then mark it as used
+        self.assert_unique_nonce(nonce)?;
+        self.assert_valid_signature(&nonce.to_be_bytes::<32>(), &signature)?;
+        self.mark_nonce_used(nonce);
```
joeykraut: can we just combine `mark_nonce_used` and `assert_unique_nonce` into one helper? like we do with the nullifier set?

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        // Refund the user's gas costs
+        transfer_eth(msg::sender(), gas_cost)?;
+
+        Ok(())
+    }
+
+    // --------------------
+    // | NAIVE SETTLEMENT |
```
joeykraut: nit: maybe "unsponsored" instead of "naive"

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        let match_result = &statement.match_result;
+        let (send_mint, send_amount) = match_result.external_party_sell_mint_amount();
+
+        // Only execute an ERC20 transfer if the input token is not the native asset
+        if !is_native_eth_address(send_mint) {
+            let send_token = IErc20::new(send_mint);
+            send_token.transfer_from(self, sender, sponsor, send_amount)?;
+        }
+
```
joeykraut, line 280: you'll need to approve the darkpool here to spend the tokens that the sponsor contract just custodied

joeykraut (approved):

contracts-stylus/src/contracts/gas_sponsor.rs
akirillo: great catch

PR #286, opened 2025-01-25
integration: tests: gas_sponsorship: unsponsored match tests

akirillo:
This PR adds the first set of integration tests for the gas sponsor contract. So far, these only test the unsponsored match paths through the gas sponsor, which are a thin wrapper around the darkpool's atomic match settlement methods, handling any necessary transfers/approvals to the gas sponsor.

### Testing
- [x] All integration tests pass

### DIFF
--- a/integration/src/tests/gas_sponsorship.rs
+++ b/integration/src/tests/gas_sponsorship.rs
@@ -1,0 +1,66 @@
+//! Integration tests for gas sponsorship
+
+use eyre::Result;
+use test_helpers::integration_test_async;
+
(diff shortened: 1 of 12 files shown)


integration/src/tests/gas_sponsorship.rs
```diff
@@ -0,0 +1,66 @@
+//! Integration tests for gas sponsorship
+
+use eyre::Result;
```
joeykraut, line 3: i like how you've set this up

joeykraut (approved):

PR #288, opened 2025-01-28
contracts-stylus: contracts: gas_sponsor: configurable refund addr + better accounting

akirillo:
This PR allows for a configurable refund address to be provided to the sponsored match methods, defaulting to `tx::origin` if the zero address is provided. Additionally, we do more granular gas accounting for the refund, inclusive of L1 gas & top-level function invocation gas costs.

This is still not a perfect measure, but we typically undercount by ~7k gas, which is well within an acceptable margin.

I have included a single integration test for a sponsored buy, asserting that the refunded amount is within our tolerance bounds, which passes. The remaining tests will come in a follow-up PR.

### DIFF
--- a/contracts-stylus/src/contracts/gas_sponsor.rs
+++ b/contracts-stylus/src/contracts/gas_sponsor.rs
@@ -145,12 +145,14 @@ impl GasSponsorContract {
     /// Sponsor the gas costs of an atomic match settlement with the caller as
     /// the receiver
     #[payable]
+    #[allow(clippy::too_many_arguments)]
     pub fn sponsor_atomic_match_settle(
         &mut self,
         internal_party_match_payload: Bytes,
         valid_match_settle_atomic_statement: Bytes,
         match_proofs: Bytes,
         match_linking_proofs: Bytes,
+        refund_address: Address,
         nonce: U256,
         signature: Bytes,
     ) -> Result<(), Vec<u8>> {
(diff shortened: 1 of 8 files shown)


contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -192,8 +222,18 @@ impl GasSponsorContract {
             );
         }
 
-        // Take note of the initial tx gas budget
-        let initial_gas = evm::gas_left();
+        // Calculate the calldata cost of invoking this method before we consume the
+        // args
+        let calldata_len = NUM_BYTES_ADDRESS
```
joeykraut: nit: aren't we missing the selector here?

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -208,9 +248,33 @@ impl GasSponsorContract {
...
 
-        // Track the total gas spent, including cost of remaining operations
-        let gas_spent = U256::from(initial_gas - evm::gas_left() + TRANSFER_GAS_COST);
-        let gas_cost = tx::gas_price() * gas_spent;
+        // Track the total gas spent, including cost of remaining operations.
+        // We frontload as many operations as possible so the final evm::gas_left()
+        // call is as accurate as possible.
+
+        let refund_address =
+            if refund_address == Address::ZERO { tx::origin() } else { refund_address };
```
joeykraut, line 263: we should probably sign the refund address and check a signature here, along with the nonce for replay protection. This prevents the refund address from being malleable if a bundle is handed to another contract that becomes aware of this pattern

akirillo: great catch

joeykraut (approved):

PR #290, opened 2025-02-01
contracts-stylus: contracts: gas_sponsor: emit amount sponsored event

akirillo:
This PR adds an `AmountSponsored` event to be emitted after successful gas sponsorship, indicating how much was refunded along with the nonce used. This can be used for accurate tracking of gas sponsorship expenditures by the sponsor.

### DIFF
--- a/contracts-stylus/src/utils/solidity.rs
+++ b/contracts-stylus/src/utils/solidity.rs
@@ -96,6 +96,7 @@ sol! {
     event InsufficientSponsorBalance(uint256 indexed nonce);
     event NonceUsed(uint256 indexed nonce);
     event GasSponsorPausedFallback(uint256 indexed nonce);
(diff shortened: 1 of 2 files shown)


contracts-stylus/src/utils/solidity.rs
```diff
@@ -96,6 +96,7 @@ sol! {
     event InsufficientSponsorBalance(uint256 indexed nonce);
     event NonceUsed(uint256 indexed nonce);
     event GasSponsorPausedFallback(uint256 indexed nonce);
+    event AmountSponsored(uint256 indexed amount, uint256 indexed nonce);
```
joeykraut: nit: I'd call this something like `SponsoredExternalMatch` or similar, naming the "event" not the quantity observed is the correct approach. the fields can describe quantities

joeykraut (approved):

PR #293, opened 2025-03-05
contracts-stylus: gas_sponsor: support gas sponsorship through buy-side token

akirillo:
This PR amends the gas sponsorship contract to support gas refunds in-kind, through the buy-side token. We also add some more accurate accounting of the invocation costs for the method.

Updates to the integration tests will come in a follow-up PR.


contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -182,6 +206,29 @@ impl GasSponsorContract {
         Ok(())
     }
 
+    /// Receives ERC20 tokens from the caller.
+    pub fn receive_tokens(&mut self, token: Address, amount: U256) -> Result<(), Vec<u8>> {
+        let token_contract = IErc20::new(token);
+        token_contract.transfer_from(Call::new(), msg::sender(), contract::address(), amount)?;
+        Ok(())
+    }
```
joeykraut: what's the purpose of this method? Won't we just be sending directly from some funding account via `IERC20.transfer`?

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -196,8 +243,10 @@ impl GasSponsorContract {
         valid_match_settle_atomic_statement: Bytes,
         match_proofs: Bytes,
         match_linking_proofs: Bytes,
-        refund_address: Address,
         nonce: U256,
+        refund_address: Address,
+        refund_native_eth: bool,
+        conversion_rate: U256,
```
joeykraut: you'll need to make a new method, this isn't backwards compatible so calls will fail for any time between the auth server upgrade and the contracts upgrade. I'd just factor out a method `sponsor_atomic_match_settle_with_options` and call that method from this one with sensible defaults.

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -224,8 +275,10 @@ impl GasSponsorContract {
         valid_match_settle_atomic_statement: Bytes,
         match_proofs: Bytes,
         match_linking_proofs: Bytes,
-        refund_address: Address,
         nonce: U256,
+        refund_address: Address,
```
joeykraut, line 300: same here, need to be backwards compatible.

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -234,95 +287,63 @@ impl GasSponsorContract {
...
-            + NUM_BYTES_SIGNATURE;
-
-        let calldata_gas = (calldata_len as u64) * GAS_PER_CALLDATA_BYTE;
-
-        // Verify the signature over the nonce + refund address,
-        // then mark the nonce as used
-        let mut message = nonce.to_be_bytes::<NUM_BYTES_U256>().to_vec();
+        // Verify the signature over the conversion rate, nonce, and refund address
+        let mut message = conversion_rate.to_be_bytes::<NUM_BYTES_U256>().to_vec();
+        message.extend_from_slice(nonce.to_be_bytes::<NUM_BYTES_U256>().as_slice());
```
joeykraut: this signature upgrade is also not backwards compatible

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -443,4 +440,275 @@ impl GasSponsorContract {
...
+        signature: signature.0.into(),
+    };
+    let calldata_len = calldata.abi_encoded_size();
+    let calldata_gas = (calldata_len as u64) * GAS_PER_CALLDATA_BYTE;
+
+    // Compute the initialization gas cost, pessimistically assuming
+    // that the gas sponsor contract is uncached.
+
+    // TODO: Check the `ArbWasmCache` precompile to see if the gas sponsor
+    // contract is cached
```
joeykraut: i think we should actually make the opposite assumption, especially if it simplifies any of this code. I think it's a more accurate assumption (the cache is currently not full), and it's probably not a huge effect on the rebate.

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -443,4 +440,275 @@ impl GasSponsorContract {
...
+    evm::log(SponsoredExternalMatch { amount: gas_cost, token: Address::ZERO, nonce });
+
+    Ok(())
+}
+
+/// Refunds the user's gas costs through the buy-side token.
+fn refund_through_buy_token(
+    refund_address: Address,
+    buy_token_addr: Address,
+    gas_cost: U256,
```
joeykraut, line 705: can we annotate here, above, and throughout the contract what the units of the gas cost and conversion rate are? Is it gwei and `base/gwei * precision`, or something else? 

Would be good in the public methods to note this so that any client work can expect a stable interface for units

joeykraut (commented):
blocking to make sure this is backwards compatible

contracts-stylus/src/contracts/gas_sponsor.rs
akirillo: Currently, our mainnet darkpool contract is uncached. We have to [periodically update our bid for inclusion in the cache](https://docs.arbitrum.io/stylus/how-tos/caching-contracts#key-features), so it's definitely possible for it to get back to this state even if once cached. Surprisingly, uncached initialization costs around 20k gas, whereas cached about 300, IIRC. I guess this still isn't a huge difference in dollar cost of the rebate, but I would like to incrementally make the gas calculation logic increasingly accurate, this is a step in that direction.

contracts-stylus/src/contracts/gas_sponsor.rs
akirillo: fair point ya

contracts-stylus/src/contracts/gas_sponsor.rs
joeykraut: Ah okay good to know, fair approach here in that case

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -227,102 +267,102 @@ impl GasSponsorContract {
...
+        signature: Bytes,
     ) -> Result<(), Vec<u8>> {
         // Take note of the initial tx gas budget
         let initial_gas = evm::gas_left();
 
         // If gas sponsorship is paused, follow through with naive settlement
         if self.is_paused()? {
             evm::log(GasSponsorPausedFallback { nonce });
-            return self.process_atomic_match_settle_with_receiver(
+            self.atomic_match_inner(
```
joeykraut: we could combine this and the call to `atomic_match_inner` below as:
```rust
self.atomic_match_inner(...)?;
if self.is_paused()? {
    return Ok(())
}
```

joeykraut (approved):

contracts-stylus/src/contracts/gas_sponsor.rs
akirillo: Will do this in the follow-up PR to simplify rebasing

PR #299, opened 2025-03-11
contracts-stylus: gas_sponsor; integration: fix receiver resolution

akirillo:
This PR ensures that the receiver address is resolved consistently for both the darkpool invocation, and the execution of the gas refund. Additionally, we add a new integration test asserting that the refund correctly goes to the resolved receiver in the case of an in-kind refund with no explicit refund address nor receiver (namely, `msg.sender`). This new test passes.


joeykraut (approved):
nice catch

PR #301, opened 2025-03-21
contracts-stylus: Upgrade `stylus-sdk` to `v0.8.3`

joeykraut:
### Purpose
This PR upgrades the `stylus-sdk` version to `v0.8.3` and makes all necessary changes. These largely involve changing method signatures and receiver types on contract methods.

### Testing
- Deferred until the nitro devnode is working again

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ ark-crypto-primitives = { version = "0.4", default-features = false, features = 
     "crh",
     "merkle_tree",
 ] }
-alloy-primitives = { version = "=0.7.7", default-features = false }
-alloy-sol-types = { version = "=0.7.7", default-features = false }
+alloy-primitives = { version = "0.8.20", default-features = false }
+alloy-sol-types = { version = "0.8.20", default-features = false }
 serde = { version = "1.0.197", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 serde_with = { version = "3.4", default-features = false, features = [
@@ -45,7 +45,7 @@ arbitrum-client = { git = "https://github.com/renegade-fi/renegade.git" }
 itertools = "0.12"
 mpc-plonk = { git = "https://github.com/renegade-fi/mpc-jellyfish.git" }
 jf-primitives = { git = "https://github.com/renegade-fi/mpc-jellyfish.git", default-features = false }
-stylus-sdk = { git = "https://github.com/joeykraut/stylus-sdk-rs" }
+stylus-sdk = { version = "0.8.3" }
 tracing = { version = "0.1", features = ["log"] }
 tracing-subscriber = "0.3"
 
--- a/contracts-stylus/src/contracts/core/core_settlement.rs
+++ b/contracts-stylus/src/contracts/core/core_settlement.rs
@@ -3,8 +3,6 @@
 //! contract and that certain storage elements are set by the outer contract. As
 //! such, its storage layout must exactly align with that of the outer contract.
 
-use core::borrow::BorrowMut;
-
 use crate::{
     assert_result,
     contracts::core::core_helpers::{call_settlement_verifier, fetch_vkeys, rotate_wallet},
@@ -38,7 +36,6 @@ use contracts_common::types::{
 use stylus_sdk::{
     abi::Bytes,
     alloy_primitives::{Address, U256},
-    msg,
     prelude::*,
     storage::{StorageAddress, StorageArray, StorageBool, StorageMap, StorageU256, StorageU64},
 };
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -45,7 +45,7 @@ arbitrum-client = { git = "https://github.com/renegade-fi/renegade.git" }
 itertools = "0.12"
 mpc-plonk = { git = "https://github.com/renegade-fi/mpc-jellyfish.git" }
 jf-primitives = { git = "https://github.com/renegade-fi/mpc-jellyfish.git", default-features = false }
-stylus-sdk = { git = "https://github.com/joeykraut/stylus-sdk-rs" }
+stylus-sdk = { version = "0.8.3" }
```
akirillo, line 48: IMO we should pin this exactly to `=0.8.3` to be safe in the face of `cargo update`s

contracts-stylus/src/contracts/core/core_settlement.rs
```diff
@@ -202,8 +199,8 @@ impl CoreSettlementContract {
     /// [`contracts_common::types::MatchProofs`] struct, and the
     /// `match_linking_proofs` argument is the serialization of the
     /// [`contracts_common::types::MatchLinkingProofs`] struct
-    pub fn process_match_settle<S: TopLevelStorage + BorrowMut<Self>>(
```
akirillo, line 205: good call removing these generics in the darkpool sub-contracts since they don't need a direct wrapper test contract. in my experience this makes navigating the codebase w/ rust-analyzer way easier

contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -375,16 +375,18 @@ impl GasSponsorContract {
...
+            send_token.approve(&mut *self, darkpool_address, send_amount)?;
+            send_token.transfer_from(&mut *self, sender, sponsor, send_amount)?;
         }
 
         // Call the darkpool contract's `process_atomic_match_settle_with_receiver`
         // method. We pass along the message value in case the input token is
         // the native asset
         let darkpool = IDarkpool::new(darkpool_address);
+        #[allow(deprecated)]
+        let ctx = CallWithValue::new().value(self.vm().msg_value());
```
akirillo, line 387: why do we use `CallWithValue` if it's deprecated? does the sdk offer no alternative?

contracts-stylus/src/contracts/test_contracts/dummy_erc20.rs
```diff
@@ -152,14 +151,15 @@ impl Erc20 {
...
+        if from != sender {
             let mut sender_allowances = self.allowances.setter(from);
-            let mut allowance = sender_allowances.setter(msg::sender());
+            let mut allowance = sender_allowances.setter(sender);
             let old_allowance = allowance.get();
             if old_allowance < value {
                 return Err(Erc20Error::InsufficientAllowance(InsufficientAllowance {
                     owner: from,
-                    spender: msg::sender(),
+                    spender: self.vm().msg_sender(),
```
akirillo, line 162: nit: could use `sender` here as well

contracts-stylus/src/utils/helpers.rs
```diff
@@ -155,6 +157,7 @@ pub fn map_calldata_ser_error<E>(_e: E) -> Vec<u8> {
     not(any(feature = "darkpool-core", feature = "darkpool-test-contract")),
     allow(dead_code)
 )]
+#[allow(deprecated)]
 pub fn static_call_helper<C: SolCall>(
```
akirillo, line 161: I take it `static_call_helper` is marked `deprecated` because it calls `stylus_sdk::call::static_call` (etc.) under the hood. why not just call `storage.vm().static_call(...)`? Same goes for `call_helper`, `delegate_call_helper`

contracts-stylus/src/utils/helpers.rs
```diff
@@ -263,6 +268,7 @@ macro_rules! if_verifying {
             use contracts_common::constants::DEVNET_CHAINID;
             use $crate::{assert_result, utils::constants::VERIFICATION_DISABLED_ERROR_MESSAGE};
 
+            #[allow(deprecated)]
```
akirillo, line 271: if `if_verifying!` becomes cumbersome to maintain we can honestly remove it IMO, we never use it in testing

contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -128,8 +126,8 @@ impl DarkpoolContract {
 
     /// Initializes the Darkpool
     #[allow(clippy::too_many_arguments)]
-    pub fn initialize<S: TopLevelStorage + BorrowMut<Self>>(
```
akirillo, line 131: Hmm... have you tried compiling with the `darkpool-test-contract` feature set? I'm getting compilation errors. This is why we pepper this `TopLevelStorage` generic throughout the `DarkpoolContract`. When compiling the `DarkpoolTestContract`, the `#[entrypoint]` isn't set on the `DarkpoolContract`, so these methods fail to compile with the error ``"the trait bound `DarkpoolContract: TopLevelStorage` is not satisfied``. The methods themselves are still linked in because the `DarkpoolTestContract` inherits from the `DarkpoolContract`

akirillo (changes requested):
Mostly lgtm but it looks like this breaks compilation of the darkpool testing contract by removing the generics in the `DarkpoolContract`

contracts-stylus/src/utils/helpers.rs
joeykraut: the old interfaces are cleaner, their new ones require a lot of cruft.

joeykraut:
going to abandon this for now

PR #302, opened 2025-03-28
contracts-stylus, scripts: Define malleable match skeleton and generate verification key

joeykraut:
### Purpose
This PR:
- Scaffolds the handlers and types for verifying a malleable atomic match settlement.
- Adds logic in the `gen-vkeys` script to generate a key for `VALID MALLEABLE MATCH SETTLE ATOMIC`

### Todo
- Contracts don't build after the circular dependency was removed from the contracts <-> relayer repos.

### Testing
- Deferred until implementation


scripts/src/commands.rs
```diff
@@ -579,15 +582,16 @@ pub async fn upgrade(
...
+    VWC: SingleProverCircuit,   // VALID WALLET CREATE
+    VWU: SingleProverCircuit,   // VALID WALLET UPDATE
+    VRFS: SingleProverCircuit,  // VALID RELAYER FEE SETTLEMENT
+    VOFS: SingleProverCircuit,  // VALID OFFLINE FEE SETTLEMENT
+    VFR: SingleProverCircuit,   // VALID FEE REDEMPTION
+    VC: SingleProverCircuit,    // VALID COMMITMENTS
+    VR: SingleProverCircuit,    // VALID REBLIND
+    VMS: SingleProverCircuit,   // VALID MATCH SETTLE
+    VMSA: SingleProverCircuit,  // VALID MATCH SETTLE ATOMIC
+    VMMSA: SingleProverCircuit, // VALID MALLEABLE MATCH SETTLE ATOMIC
```
akirillo, line 594: i'm so sorry for this method

akirillo (approved):

scripts/src/commands.rs
joeykraut: There are worse sins in the world of rust generics

PR #304, opened 2025-03-31
contracts-core, contracts-utils: Remove `ethers` dependency

joeykraut:
### Purpose
This PR removes the `ethers` dependency in `contracts-core` and `contracts-utils` in favor of `alloy`. This begins a larger transition off of `ethers` in the stack.

### Todo
- Migrate the rest of the contacts repo
- Fix integration tests once all crates are migrated

### Testing
- [x] Unit tests pass

### DIFF
--- a/contracts-utils/src/crypto.rs
+++ b/contracts-utils/src/crypto.rs
@@ -1,13 +1,13 @@
 //! Helpful cryptographic utilities
 
+use alloy::{
+    primitives::{PrimitiveSignature, U256},
+    signers::k256::ecdsa::SigningKey,
+};
+use alloy_primitives::keccak256;
 use circuit_types::keychain::PublicSigningKey as CircuitPubkey;
 use contracts_common::{
     backends::HashBackend, constants::HASH_OUTPUT_SIZE, types::PublicSigningKey,
-};
-use ethers::{
-    core::k256::ecdsa::SigningKey,
-    types::{Signature, U256},
-    utils::keccak256,
 };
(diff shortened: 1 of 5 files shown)


contracts-utils/src/crypto.rs
```diff
@@ -19,7 +19,7 @@ pub struct NativeHasher;
 
 impl HashBackend for NativeHasher {
     fn hash(input: &[u8]) -> [u8; HASH_OUTPUT_SIZE] {
-        keccak256(input)
+        *keccak256(input)
```
akirillo, line 22: one thing I would be careful about here: `alloy`'s `keccak256` may [try to invoke a native VM hook for Keccak 256](https://github.com/alloy-rs/core/blob/main/crates/primitives/src/utils/mod.rs#L148) if the right Cargo features aren't set, which will result in runtime panics due to calling missing symbols.

The `native-keccak` feature which triggers this behavior isn't on by default but may get toggled on by the `stylus-sdk` crate. We only use the structs/methods in this file for testing helpers so we want to be sure to enable the `tiny-keccak` feature on `alloy` here to avoid this issue.

akirillo (approved):
just one gotcha i've dealt with before, o/w lgtm

contracts-utils/src/crypto.rs
joeykraut: I'll look out for this, but the SDK does not currently enable it, as per `cargo tree`.

PR #305, opened 2025-03-31
scripts: Remove `ethers` dependency

joeykraut:
### Purpose
This PR removes the `ethers` dependency from the `scripts` package, and replaces the helpers with `alloy` equivalents.

### Todo
- Remove `ethers` from the integration tests and ensure they pass

### Testing
- [x] Scripts build and run

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ ark-crypto-primitives = { version = "0.4", default-features = false, features = 
     "crh",
     "merkle_tree",
 ] }
+alloy = { version = "0.12", default-features = false }
+alloy-contract = { version = "=0.12", default-features = false }
 alloy-primitives = { version = "=0.7.7", default-features = false }
 alloy-sol-types = { version = "=0.7.7", default-features = false }
 serde = { version = "1.0.197", default-features = false, features = ["derive"] }
--- a/contracts-utils/src/crypto.rs
+++ b/contracts-utils/src/crypto.rs
@@ -39,7 +39,7 @@ pub fn random_keypair<R: CryptoRng + RngCore>(rng: &mut R) -> (SigningKey, Publi
 /// key, as expected in ECDSA
 pub fn hash_and_sign_message(signing_key: &SigningKey, msg: &[u8]) -> PrimitiveSignature {
     let msg_hash = keccak256(msg);
-    let (sig, recovery_id) = signing_key.sign_prehash_recoverable(&msg_hash.as_slice()).unwrap();
+    let (sig, recovery_id) = signing_key.sign_prehash_recoverable(msg_hash.as_slice()).unwrap();
     let r: U256 = U256::from_be_bytes(sig.r().to_bytes().into());
     let s: U256 = U256::from_be_bytes(sig.s().to_bytes().into());
 
--- a/scripts/src/commands.rs
+++ b/scripts/src/commands.rs
@@ -1,6 +1,9 @@
 //! Implementations of the various deploy scripts
 
-use alloy_primitives::U256;
+use alloy::{
+    primitives::{Address, Bytes, U256},
+    providers::Provider,
+};
 use circuit_types::traits::SingleProverCircuit;
 use circuits::zk_circuits::{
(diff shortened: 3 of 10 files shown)


contracts-utils/src/crypto.rs
```diff
@@ -39,7 +39,7 @@ pub fn random_keypair<R: CryptoRng + RngCore>(rng: &mut R) -> (SigningKey, Publi
 /// key, as expected in ECDSA
 pub fn hash_and_sign_message(signing_key: &SigningKey, msg: &[u8]) -> PrimitiveSignature {
     let msg_hash = keccak256(msg);
-    let (sig, recovery_id) = signing_key.sign_prehash_recoverable(&msg_hash.as_slice()).unwrap();
+    let (sig, recovery_id) = signing_key.sign_prehash_recoverable(msg_hash.as_slice()).unwrap();
```
akirillo, line 42: this was in the last PR, but i just came across it in my own testing:

returning a `PrimitiveSignature` as below is fine, but if you call `as_bytes()` on that signature, it will serialize the `v` value as `27` or `28`, not as `0` or `1`. The latter is the previous behavior and what is expected by the contracts.

so we should either:
1. change the contracts to expect `27` / `28` as the `v` value in the signature, or
2. be sure to subtract `27` from the final byte of the signature

Cargo.toml
```diff
@@ -19,6 +19,8 @@ ark-crypto-primitives = { version = "0.4", default-features = false, features =
     "crh",
     "merkle_tree",
 ] }
+alloy = { version = "0.12", default-features = false }
+alloy-contract = { version = "=0.12", default-features = false }
 alloy-primitives = { version = "=0.7.7", default-features = false }
 alloy-sol-types = { version = "=0.7.7", default-features = false }
```
akirillo, line 25: perhaps out of scope for this wave of "remove `ethers`" PRs, but we can start bumping alloy core deps to their latest versions now, yeah?

scripts/src/commands.rs
```diff
@@ -348,50 +344,39 @@ pub async fn deploy_darkpool_proxy(
...
 ) -> Result<(), ScriptError> {
-    // Get proxy contract ABI and bytecode
-    let abi: Contract =
-        serde_json::from_str(PROXY_ABI).map_err(|e| ScriptError::ArtifactParsing(e.to_string()))?;
-
-    let bytecode =
-        Bytes::from_hex(PROXY_BYTECODE).map_err(|e| ScriptError::ArtifactParsing(e.to_string()))?;
-
-    let proxy_factory = ContractFactory::new(abi, bytecode, client.clone());
-
```
akirillo, line 365: fantastic

scripts/src/commands.rs
```diff
@@ -348,50 +344,39 @@ pub async fn deploy_darkpool_proxy(
...
-    );
+    let slot = U256::from_str(PROXY_ADMIN_STORAGE_SLOT).unwrap();
+    let slot_value = client
+        .provider()
+        .get_storage_at(proxy_address, slot)
+        .await
+        .map_err(|e| ScriptError::ContractInteraction(e.to_string()))?;
+    let addr_bytes = &slot_value.to_be_bytes_vec()
+        [NUM_BYTES_STORAGE_SLOT - NUM_BYTES_ADDRESS..NUM_BYTES_STORAGE_SLOT];
+    let proxy_admin_address = Address::from_slice(addr_bytes);
```
akirillo, line 379: now that we're using alloy, I think we should be able to use [`Address::from_word`](https://docs.rs/alloy/latest/alloy/primitives/struct.Address.html#method.from_word)

akirillo (approved):

Cargo.toml
joeykraut: no, it's constrained by the stylus sdk

contracts-utils/src/crypto.rs
joeykraut: i'll just do what i did in the solidity implementation [here](https://github.com/renegade-fi/renegade-solidity-contracts/blob/459f7aff3252b6ba10e36a1291e671426b9b09dc/src/libraries/darkpool/WalletOperations.sol#L178-L181), we really shouldn't be so sensitive to client impls

PR #307, opened 2025-04-03
contracts-stylus: core: core-settlement: Implement malleable match settlement

joeykraut:
### Purpose
This PR implements the core logic for settling a malleable match minus the following:
- Updating the internal party's shares
- Validating the malleable match proof
Those will come in a follow-up

### Testing
- [x] Deferred to follow-up

### DIFF
--- a/contracts-common/src/types.rs
+++ b/contracts-common/src/types.rs
@@ -9,7 +9,8 @@ use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
 use crate::{
-    constants::{NUM_SELECTORS, NUM_U64S_FELT, NUM_WIRE_TYPES},
+    constants::{FIXED_POINT_PRECISION_BITS, NUM_SELECTORS, NUM_U64S_FELT, NUM_WIRE_TYPES},
+    custom_serde::scalar_to_u256,
     serde_def_types::*,
 };
 
@@ -31,6 +32,9 @@ pub type G2BaseField = Fq2;
 /// Type alias for a 256-bit prime field element in Montgomery form
 pub type MontFp256<P> = Fp256<MontBackend<P, NUM_U64S_FELT>>;
 
+/// The revert message when an invalid base amount is provided
+pub const ERROR_INVALID_BASE_AMT: &[u8] = b"invalid base amount";
+
 /// A fixed-point representation of a real number
 ///
 /// In the Renegade darkpool, a fixed point representation of a real number `r`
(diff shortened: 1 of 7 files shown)


contracts-common/src/types.rs
```diff
@@ -40,7 +44,25 @@ pub type MontFp256<P> = Fp256<MontBackend<P, NUM_U64S_FELT>>;
...
+    /// Computes `(self.repr * scalar) / 2^FIXED_POINT_PRECISION_BITS`
+    ///
+    /// The repr already has the fixed point scaling value, so we only need to
+    /// undo the scaling once to get the desired result. Because division
+    /// naturally truncates, this will implement the floor of the above
+    /// division.
+    ///
+    /// # Warning
+    /// This function is unsafe because it does not check for overflows
+    pub fn unsafe_fixed_point_mul(&self, scalar: U256) -> U256 {
```
akirillo, line 62: i'd suggest wrapping this in an `unsafe`, but i guess the method title is clear enough

contracts-common/src/types.rs
```diff
@@ -418,14 +440,26 @@ impl FeeTake {
...
     /// The fee rate for the protocol
     pub protocol_fee_rate: FixedPoint,
 }
 
+impl FeeRates {
+    /// Get a fee take from the fee rates given a receive amount
+    pub fn get_fee_take(&self, receive_amount: U256) -> FeeTake {
+        // SAFETY: The fee rates are constrained in-circuit to be less than 2^63, and
+        // the receive amount is constrained to be less than 2^100, so the
+        // product is less than 2^163, which fits in a uint256
```
akirillo, line 456: very helpful doc comment, thank you

contracts-common/src/types.rs
```diff
@@ -502,6 +535,39 @@ pub struct BoundedMatchResult {
...
+        // Validate the match amount
+        let amount_too_low = base_amount < self.min_base_amount;
+        let amount_too_high = base_amount > self.max_base_amount;
+        if amount_too_low || amount_too_high {
+            return Err(ERROR_INVALID_BASE_AMT.into());
+        }
+
+        // Compute the quote amount
+        let price = self.price;
+        let quote_amount = price.unsafe_fixed_point_mul(base_amount);
```
akirillo, line 558: we're sure this is safe? i'm sure `base_amount` is constrained to be < 2^100 in-circuit, but i don't recall on the price

contracts-stylus/src/contracts/core/core_settlement.rs
```diff
@@ -374,7 +374,65 @@ impl CoreSettlementContract {
...
+                storage.borrow_mut().external_match_protocol_fee(bounded_match_result.base_mint);
+            let protocol_fee = u256_to_scalar(protocol_fee_u256)?;
+            assert_result!(
+                internal_fee_rates.protocol_fee_rate.repr == protocol_fee,
+                INVALID_PROTOCOL_FEE_ERROR_MESSAGE
+            )?;
+            assert_result!(
+                external_fee_rates.protocol_fee_rate.repr == protocol_fee,
+                INVALID_PROTOCOL_FEE_ERROR_MESSAGE
+            )?;
```
akirillo, line 409: if both the internal & external protocol fee rate must be the same, why are they separate statement vars? for down-the-line flexibility?

akirillo (approved):

contracts-common/src/types.rs
joeykraut: well it's not really unsafe in the rust sense

contracts-common/src/types.rs
joeykraut: prices have validated bitlength

contracts-stylus/src/contracts/core/core_settlement.rs
joeykraut: yes

PR #310, opened 2025-04-04
contracts-stylus: core-settlement: Apply malleable match to wallet shares

joeykraut:
### Purpose
This PR applies the derived external match directly to the internal party's wallet shares when processing a malleable match. This is done by deserializing the shares into a runtime type, applying the update, then re-serializing for wallet rotation.

### Todo
- Split apart core settlement contract, it now is above the bytecode size limit.

### Testing
- Deferred until follow up

### DIFF
--- a/contracts-common/src/types/wallet.rs
+++ b/contracts-common/src/types/wallet.rs
@@ -1,0 +1,234 @@
+//! Types related to darkpool wallets
+
+use super::{
+    u256_to_scalar, ExternalMatchResult, FeeRates, FixedPoint, OrderSettlementIndices,
+    PublicEncryptionKey, PublicKeychain, ScalarField,
+};
+use crate::types::{BabyJubJubPoint, PublicIdentificationKey, PublicRootKey};
+use alloc::vec::Vec;
+
+/// Error message for when a vector cannot be sized into an array
+pub const VEC_SIZE_ERROR: &[u8] = b"Vector length does not match array size";
+
+/// The maximum number of balances in a wallet
+pub const MAX_BALANCES: usize = 10;
+/// The maximum number of orders in a wallet
+pub const MAX_ORDERS: usize = 4;
+/// The number of scalars in a serialized wallet share
+pub const NUM_WALLET_SCALARS: usize = 70;
+
+/// Helper function to size a vector into an array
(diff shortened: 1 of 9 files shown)


contracts-common/src/types/wallet.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// match fees
+    pub managing_cluster: PublicEncryptionKey,
+    /// The public share of the wallet's blinder
+    pub blinder: ScalarField,
+}
+
+#[cfg(feature = "core-settlement")]
+impl WalletShare {
+    /// Apply an external match to a wallet's shares
+    pub fn apply_external_match_to_shares(
```
akirillo, line 49: all the arithmetic in this method is unchecked. in practice this is fine since the method is applied after a proof containing these inputs is verified, so the arithmetic is effectively checked in-circuit (or, the outer bounds of the potential arithmetic operations are checked, and the values are constrained to be within bounds).

but still, I'd feel more comfortable if this were documented on the method, or even if it was called `unsafe_apply_external_match_to_shares` or similar

contracts-common/src/types/wallet.rs
```diff
@@ -0,0 +1,231 @@
...
+        recv_bal.relayer_fee_balance += u256_to_scalar(internal_party_fees.relayer_fee).unwrap();
+        recv_bal.protocol_fee_balance += u256_to_scalar(internal_party_fees.protocol_fee).unwrap();
+
+        // Deduct the send amount from the wallet's balances
+        let send_bal = &mut self.balances[indices.balance_send as usize];
+        send_bal.amount -= u256_to_scalar(send_amount).unwrap();
+    }
+
+    /// Serialize a wallet share into a list of scalars
+    pub fn scalar_serialize(&self) -> [ScalarField; NUM_WALLET_SCALARS] {
```
akirillo, line 78: why not have `scalar_serialize` be defined in an `impl ScalarSerializable` block? p much 0 overhead/complexity for something that may be useful in the future

akirillo (approved):

contracts-common/src/types/wallet.rs
joeykraut: I'll document it, this doesn't make sense to call it unsafe though

contracts-common/src/types/wallet.rs
joeykraut: because I want to size the output

PR #311, opened 2025-04-04
contracts-stylus: darkpool: Transition to delegate callee mapping

joeykraut:
### Purpose
This PR removes the inlined delegate callee addresses from contract storage (replacing them with storage gaps), and instead uses a mapping from a delegate callee "selector" to its address. 

This will require special logic when upgraded to initialize these addresses.

### Testing
- [x] All unit and integration tests pass

### DIFF
--- a/contracts-stylus/src/contracts/darkpool.rs
+++ b/contracts-stylus/src/contracts/darkpool.rs
@@ -33,7 +33,8 @@ use crate::{
             VerifierSettlementAddressChanged, VkeysAddressChanged,
         },
     },
-    CORE_SETTLEMENT_DELEGATE_SELECTOR, CORE_WALLET_OPS_DELEGATE_SELECTOR, MERKLE_DELEGATE_SELECTOR,
+    CORE_SETTLEMENT_DELEGATE_SELECTOR, CORE_WALLET_OPS_DELEGATE_SELECTOR,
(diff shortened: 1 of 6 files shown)


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -391,17 +386,31 @@ impl DarkpoolContract {
...
+    ) -> Result<(), Vec<u8>> {
+        DarkpoolContract::_check_owner(storage)?;
+        check_address_not_zero(address)?;
+        storage.borrow_mut().delegate_addresses.insert(selector, address);
+
+        Ok(())
+    }
+
     /// Sets the darkpool core address
     pub fn set_core_wallet_ops_address<S: TopLevelStorage + BorrowMut<Self>>(
```
akirillo, line 403: do we want to keep these `set_*_address` methods around, now that we have `set_delegate_address`? they feel vestigial and may become obsolete anyway as we continue to split up contracts

akirillo (approved):

contracts-stylus/src/contracts/darkpool.rs
joeykraut: yes, I don't want a blanket `set_delegate_address` method on the interface being used. Too prone to error given that we're using selectors for it

PR #312, opened 2025-04-05
meta: patch stylus SDK to latest version

akirillo:
This PR updates the Stylus SDK to the latest version, making the necessary changes to have all crates in the project compile.

Two changes of note:
1. We remove the now-unused methods on the gas sponsor contract. They are not worth maintaining in a change like this, and I will need to update the integration tests later anyway.
2. A lot of our current usage of the Stylus SDK invokes deprecated methods. We leave these in for now to mitigate the blast radius of this PR.

### TODO:
- [x] Replace deprecated Stylus SDK methods where appropriate
- [x] Update integration tests & get them passing

### DIFF
--- a/integration/Cargo.toml
+++ b/integration/Cargo.toml
@@ -34,6 +34,6 @@ inventory = "0.3"
 num-bigint = "0.4"
 ruint = { workspace = true }
 
-test-helpers = { git = "https://github.com/renegade-fi/renegade.git" }
-renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+test-helpers = { path = "../../renegade/test-helpers" }
+renegade-crypto = { workspace = true, default-features = false }
 colored = "2"
--- a/integration/Cargo.toml
+++ b/integration/Cargo.toml
@@ -34,6 +34,6 @@ inventory = "0.3"
 num-bigint = "0.4"
 ruint = { workspace = true }
 
-test-helpers = { path = "../../renegade/test-helpers" }
+test-helpers = { git = "https://github.com/renegade-fi/renegade.git" }
 renegade-crypto = { workspace = true, default-features = false }
 colored = "2"
--- a/scripts/src/commands.rs
+++ b/scripts/src/commands.rs
@@ -1,9 +1,12 @@
 //! Implementations of the various deploy scripts
 
 use alloy::{
+    network::TransactionBuilder,
     primitives::{Address, Bytes, U256},
     providers::Provider,
+    rpc::types::TransactionRequest,
 };
+use alloy_sol_types::SolConstructor;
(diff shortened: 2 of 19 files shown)


integration/Cargo.toml
```diff
@@ -34,6 +34,6 @@ inventory = "0.3"
 num-bigint = "0.4"
 ruint = { workspace = true }
 
-test-helpers = { git = "https://github.com/renegade-fi/renegade.git" }
-renegade-crypto = { git = "https://github.com/renegade-fi/renegade.git", default-features = false }
+test-helpers = { git = "https://github.com/renegade-fi/renegade.git", branch = "andrew/downgrade-alloy" }
```
joeykraut: reminder to remove all these branches

scripts/src/constants.rs
```diff
@@ -7,22 +7,17 @@ use alloy::sol;
...
-sol! {
-    #[allow(missing_docs)]
-    #[sol(bytecode = include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/artifacts/Permit2.bin")))]
-    #[sol(rpc)]
-    Permit2,
-    "artifacts/Permit2.abi"
-}
+/// The bytecode of the `Permit2` contract
+pub const PERMIT2_BYTECODE: &str =
+    include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/artifacts/Permit2.bin"));
```
joeykraut, line 20: why did we remove this?

scripts/src/commands.rs
```diff
@@ -422,16 +419,53 @@ pub async fn deploy_permit2(
...
+
+    let constructor_calldata = TransparentUpgradeableProxy::constructorCall {
+        _logic: implementation_address,
+        initialOwner: owner_address,
+        _data: initialization_calldata,
+    }
+    .abi_encode();
+
+    let deploy_code = [&bytecode[..], &constructor_calldata[..]].concat();
+    Ok(deploy_code)
```
joeykraut, line 449: this is very low level, alloy really provides no better way to construct contracts from bytecode? what about the helpers I had in place?

joeykraut (approved):

scripts/src/constants.rs
akirillo: Because we don't use it anywhere. We only constructed the contract instance so we could call `.address()` on it after deploying, but we get the deployed address from the transaction receipt now.

scripts/src/commands.rs
akirillo: Not for the alloy versions we have to use. We can't construct contract instances in the nice way you had set up because https://github.com/alloy-rs/core/pull/898 wasn't merged yet in that version. so the only other way to include the bytecode without literally inlining it as a string literal is this.

PR #313, opened 2025-04-05
contracts-stylus: core: Split core-settlement into three contracts

joeykraut:
### Purpose
This PR splits the `core-settlement` contract into three separate contracts for the three separate settlement method, effectively introducing a per-method split paradigm. This allows each contract to fit into the bytecode limit.

I also updated the scripts and tests to reflect this split.

### Testing
- [x] All unit and integration tests pass


contracts-stylus/src/contracts/darkpool.rs
```diff
@@ -132,7 +130,9 @@ impl DarkpoolContract {
     pub fn initialize<S: TopLevelStorage + BorrowMut<Self>>(
```
akirillo, line 130: since we'll need to re-initialize the contract when we land this upgrade via an `upgradeAndCall`, we should bump the version number to `2` in the `DarkpoolContract::_initialize` call at the bottom of this method

akirillo (approved):

PR #317, opened 2025-04-07
integration: tests: malleable-atomic-settlement: Finish test suite

joeykraut:
### Purpose
This PR adds the full test suite for malleable matches to the integration tests. This tests all valid/invalid cases as in the solidity contracts.

### Testing
- [x] All integration tests pass

### DIFF
--- a/integration/src/tests/malleable_atomic_settlement.rs
+++ b/integration/src/tests/malleable_atomic_settlement.rs
@@ -1,20 +1,173 @@
 //! Integration tests for malleable atomic settlement
+use alloy::rpc::types::TransactionReceipt;
+use alloy_primitives::{Address, U256};
+use constants::ScalarField;
+use contracts_common::types::{
+    MatchAtomicLinkingProofs, MatchAtomicProofs, MatchPayload,
+    ValidMalleableMatchSettleAtomicStatement,
+};
 use eyre::Result;
+use rand::{thread_rng, Rng};
 use scripts::utils::send_tx;
-use test_helpers::integration_test_async;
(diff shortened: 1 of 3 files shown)


integration/src/tests/malleable_atomic_settlement.rs
```diff
@@ -1,20 +1,173 @@
...
+    Ok(())
+}
+integration_test_async!(test_malleable_match__non_sender_receiver_native);
+
+// --- Invalid Test Cases --- //
+
+/// Test a malleable match on the native asset with an invalid ETH value (i.e.
+/// greater than zero)
+#[allow(non_snake_case)]
+async fn test_malleable_match__non_native_invalid_value(ctx: TestContext) -> Result<()> {
```
akirillo, line 110: A little confused about this test case. The function name implies we're testing that a non-zero `value` cannot be included in any matches that are not on the native asset. But the doc comment and `setup_malleable_match_test(true /* is_native */, ..)` imply that we're testing the case where we are trading the native asset. If the latter is the case, then this test will be flaky, since it's only expected to fail when a `value` is provided for a native asset _buy_. But `setup_malleable_match_test` generates a buy/sell with equal probability.

akirillo (approved):
mostly lgtm other than `test_malleable_match__non_native_invalid_value`, approving to unblock

integration/src/tests/malleable_atomic_settlement.rs
joeykraut: fixed, name was correct implementation was wrong.

PR #318, opened 2025-04-07
contracts-stylus: gas_sponsor: return exact output amount

akirillo:
This PR updates the gas sponsor contract's `sponsor_atomic_match_settle_with_refund_options` method to return the exact output amount received by the external party. We also emit a new log, `SponsoredExternalMatchOutput`, containing this value. This log is necessary to view the output amount outside of the EVM.

### Testing
We add a new set of integration tests asserting that the output amount is correctly calculated, all of which pass. We also add a test case asserting that a match succeeds, but no refund is issued, for in-kind refunds when the gas sponsor lacks the requisite token balance.

### DIFF
--- a/contracts-common/src/types/fees.rs
+++ b/contracts-common/src/types/fees.rs
@@ -19,7 +19,7 @@ pub struct FeeTake {
     pub protocol_fee: U256,
 }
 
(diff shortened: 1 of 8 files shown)


akirillo:
Closed in favor of #319 and #320

PR #319, opened 2025-04-07
contracts-stylus: core_atomic_match_settle, gas_sponsor: return amount received

akirillo:
This PR updates the core atomic match settlement and gas sponsor contracts to return the exact amount received by the external party in the match.

Integration tests will come in a follow-up PR to mitigate the scope of review

### DIFF
--- a/contracts-stylus/src/contracts/core/core_atomic_match_settle.rs
+++ b/contracts-stylus/src/contracts/core/core_atomic_match_settle.rs
@@ -19,7 +19,7 @@ use crate::{
             deserialize_from_calldata, get_weth_address, is_native_eth_address, postcard_serialize,
             serialize_atomic_match_statements_for_verification,
         },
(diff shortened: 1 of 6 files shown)


contracts-stylus/src/contracts/core/core_atomic_match_settle.rs
```diff
@@ -241,7 +243,15 @@ impl CoreAtomicMatchSettleContract {
         let relayer_fee_address = valid_match_settle_atomic_statement.relayer_fee_address;
         execute_atomic_match_transfers(self, receiver, fees, match_result, relayer_fee_address)?;
 
-        Ok(())
+        // Calculate the exact amount received by the external party
```
joeykraut: probably best to just have the `execute_atomic_match_transfers` method return this value rather than recomputing it.

joeykraut (approved):

PR #338, opened 2025-09-01
integration: gas-sponsorship: Allow zero gas sponsorship

joeykraut:
### Purpose
This PR explicitly returns early if the configured refund amount in the gas sponsorship contract is zero. We return immediately after settling the trade.

### Testing
- [x] Added integration tests for both native eth and in-kind gas sponsorship.


contracts-stylus/src/contracts/gas_sponsor.rs
```diff
@@ -217,8 +217,9 @@ impl GasSponsorContract {
             match_linking_proofs,
         )?;
 
-        // If gas sponsorship is paused, return early, no refunding will be done
-        if self.is_paused()? {
+        // If gas sponsorship is paused or the refund amount is zero, return early. No
+        // refunding will be done
+        if self.is_paused()? || refund_amount.is_zero() {
             log(self.vm(), GasSponsorPausedFallback { nonce });
```
akirillo: can we change the event ABI to be one that properly represents the early return condition here? something like `GasSponsorshipSkipped { nonce }`? we don't rely on the event anywhere in our own infrastructure, nor do any users that I know of, so changing it should be fine

akirillo (approved):
good catch

contracts-stylus/src/contracts/gas_sponsor.rs
joeykraut: yes will do

